PR #74, opened 2023-12-01
Fix Class "Str" not found error on fresh install

WillyReyno:
Hello,

On a fresh install, going to `/pulse` results in the following error: https://flareapp.io/share/87nK4ry5

Adding missing use for `Illuminate\Support\Str` in 3 blade views fixed the issue.

### DIFF
--- a/resources/views/livewire/cache.blade.php
+++ b/resources/views/livewire/cache.blade.php
@@ -1,3 +1,6 @@
+@php
+    use Illuminate\Support\Str;
+@endphp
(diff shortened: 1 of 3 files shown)


simonhamp:
Doesn't this mean just you don't have the aliases set up in your `config/app.php`?

CLEMREP:
After a fresh install, I get the following error :

`Call to undefined method Illuminate\Support\Benchmark::value()`

PR #80, opened 2023-12-01
[1.x] Allow ignoring route registration

toitzi:
Similiar to other first pary packages i thought it would be great to allow ignoring route registration, in order to be able to register the routes yourself. A more practical use case might be if you use something like tenancy for laravel where registring the routes only on central domains, etc.. can be beneficial and easir to be created that way

### DIFF
--- a/src/Pulse.php
+++ b/src/Pulse.php
@@ -83,6 +83,11 @@ class Pulse
     protected bool $runsMigrations = true;
 
     /**
+     * Indicates if Pulse routes will be registered.
+     */
+    protected bool $registersRoutes = true;
+
+    /**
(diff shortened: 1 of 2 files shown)


timacdonald (approved):
Tested on a fresh application and a real application. Works as expected.

Only caveat is that this would generally need to be done in the service provider's `register` method, not it's `boot` method.

taylorotwell:
Has conflicts.

toitzi:
> Has conflicts.

Resolved, there seems to be an issue with the static code analysis from the 1.x branch right now

timacdonald:
rebased this one against a fix version

PR #81, opened 2023-12-01
[1.x] Make migrations publishable

toitzi:
Similiar to other first pary packages i thought it would be great to allow publishing migrations. Same as in this pr #80: a more practical use case might be if you use something like tenancy for laravel where you might want to have those migrations for your tenant databases as well

### DIFF
--- a/src/Facades/Pulse.php
+++ b/src/Facades/Pulse.php
@@ -26,8 +26,6 @@ use Illuminate\Support\Facades\Facade;
  * @method static \Laravel\Pulse\Pulse rememberUser(\Illuminate\Contracts\Auth\Authenticatable $user)
  * @method static string css()
  * @method static string js()
- * @method static bool runsMigrations()
- * @method static \Laravel\Pulse\Pulse ignoreMigrations()
  * @method static bool registersRoutes()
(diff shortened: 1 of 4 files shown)


SRWieZ:
This will also be useful for skipping the migration in test

Example, by wrapping the `up()` function with 
```php
if (app()->runningUnitTests()) 
```

timacdonald:
@SRWieZ If you want to disable Pulse during tests, you should set the env - otherwise it'll be doing work during the test still - but just swallowing exceptions.

```xml
<server name="PULSE_ENABLED" value="false"/>
```

taylorotwell:
@timacdonald @jessarcher we are moving most of our first party packages to *only* publishing migrations (never running them behind the scenes). We are making this change for Sanctum and maybe others in Laravel 11. 

It may be good to go ahead and just take this approach here?

timacdonald:
@taylorotwell dig it. Updated this one and updated the docs over here: https://github.com/laravel/docs/pull/9176

taylorotwell:
Conflicts. 😅

jbrooksuk:
@taylorotwell I've resolved the conflicts.

PR #88, opened 2023-12-01
Manually inject livewire assets

### LINKED ISSUE #87: Pulse not loading when Livewire's `inject_assets` is `false`
### Pulse Version

v1.0.0-beta2

### Laravel Version

10.34.2

### PHP Version

8.2.4

### Description

When you're [manually including Livewire's frontend assets](https://livewire.laravel.com/docs/installation#manually-including-livewires-frontend-assets) and set `inject_assets` to `false` in Livewire's config, Pulse won't inject the required assets and won't load.

### Steps To Reproduce

Publish Livewire's config and set `inject_assets` to `false`.

toitzi:
Fixes #87 by manually adding the styles and js for livewire. This should be non breaking, and add support for those cases

### DIFF
--- a/resources/views/components/pulse.blade.php
+++ b/resources/views/components/pulse.blade.php
@@ -14,6 +14,7 @@
         <style>
             {!! Laravel\Pulse\Facades\Pulse::css() !!}
         </style>
+
         @livewireStyles
 
         <script>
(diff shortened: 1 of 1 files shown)


SRWieZ:
Ahah, you're faster than me. I got distracted by some nice thing in the code.

However, I believe Scripts should be at the end of the body as per:
https://livewire.laravel.com/docs/installation#manually-including-livewires-frontend-assets

gofish543:
Anyway we can get a nonce included on these livewire scripts? @toitzi @SRWieZ

PR #93, opened 2023-12-01
[1.x] Introducing Soft Scrollbar

devajmeireles:
This PR aims to introduce a soft and styled scroll bar:

Before:

CleanShot 2023-12-01 at 19 46 30

After:

CleanShot 2023-12-01 at 19 46 48


jessarcher:
Interesting - I haven't seen that scrollbar on Linux or MacOS. What browser/OS are you using?

devajmeireles:
> Interesting - I haven't seen that scrollbar on Linux or MacOS. What browser/OS are you using?

Chrome on MacOS.

kaiserkiwi:
You have to enable permanent visible scrollbars for the in system settings.

timacdonald:
I've just tested on Mac and can only get this to happen if I've specifically opted into permanent visible scrollbars, as @kaiserkiwi mentioned.

I don't think we should override user preferences.

This could be an accessibility concern.

PR #113, opened 2023-12-02
[1.x] Add possibility to change cache duration

pierrepavlov:
While creating custom cards you might want to be able to set a longer cache duration without having to create a custom trait.

### DIFF
--- a/src/Livewire/Concerns/RemembersQueries.php
+++ b/src/Livewire/Concerns/RemembersQueries.php
@@ -15,9 +15,9 @@ trait RemembersQueries
      *
      * @return array{0: mixed, 1: float, 2: string}
      */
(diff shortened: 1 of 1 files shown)


src/Livewire/Concerns/RemembersQueries.php
```diff
@@ -15,9 +15,9 @@ trait RemembersQueries
      *
      * @return array{0: mixed, 1: float, 2: string}
      */
-    public function remember(callable $query, string $key = ''): array
+    public function remember(callable $query, string $key = '', int $ttl = 5): array
```
timacdonald, line 18: Would be nice if this supported all the types that `Cahce::remember` supports for the TTL, not just `int`

timacdonald (approved):

PR #118, opened 2023-12-02
Allow recorder inheritance

danwall:
When trying to build a custom recorders which all inherited from a base class I ran into some issues where the config object was resolving values from the base class name, causing the recorder not to run.

Technically only the Concerns need to be updated to fix this, but I've gone ahead and updated the built-in recorders as well so that they can be extended and have their own configs.


timacdonald:
I don’t think the recorders should be updated at this point, only the concerns.

danwall:
@timacdonald thanks for the feedback, have removed those extra changes now

PR #126, opened 2023-12-04
[1.x] Improve test configuration and bump min dependencies

timacdonald:
Now tests against:

- PHP 8.1
- `--prefer-lowest`


composer.json
```diff
@@ -18,20 +18,20 @@
...
+        "illuminate/contracts": "^10.21",
+        "illuminate/database": "^10.21",
+        "illuminate/events": "^10.21",
+        "illuminate/http": "^10.21",
+        "illuminate/queue": "^10.21",
+        "illuminate/redis": "^10.21",
+        "illuminate/routing": "^10.21",
+        "illuminate/support": "^10.21",
+        "illuminate/view": "^10.21",
+        "livewire/livewire": "^3.02",
```
timacdonald, line 34: There is bug for lazy components that was fixed in `3.02`

PR #158, opened 2023-12-06
[1.x] Add ability to cancel `pulse:check` command if needed

crynobone:
While you can use `pulse:restart` it would be useful to be able to cancel `pulse:check` directly using `InteractsWithSignals`: https://laravel.com/docs/10.x/artisan#signal-handling

### DIFF
--- a/src/Commands/CheckCommand.php
+++ b/src/Commands/CheckCommand.php
@@ -5,8 +5,10 @@ namespace Laravel\Pulse\Commands;
 use Carbon\CarbonImmutable;
 use Carbon\CarbonInterval;
 use Illuminate\Console\Command;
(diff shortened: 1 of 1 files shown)


timacdonald:
@crynobone what problem does this solve? Wouldn't sending a signal, such as `ctrl-c`, already kill the command like any other Daemon, such as `queue:work`?

Setting this cache key would also restart the `pulse:work` command, so it has side-effects outside of the check command which might be weird.

crynobone:
@timacdonald look like you are correct. The issue seems to only exist with `php vendor/bin/testbench pulse:check`

timacdonald:
Ahhh, gotcha!

PR #159, opened 2023-12-06
[1.x] Fix number rounding on cache card

timacdonald:
Due to how we are handing rounding, we can sometimes show a `100%` hit rate for cache keys, even when there are misses.

This PR removes the use of `round` and just uses `int` casting to trim the decimal without performing any rounding. We maintain 2 decimal places.

### DIFF
--- a/tests/Feature/Livewire/CacheTest.php
+++ b/tests/Feature/Livewire/CacheTest.php
@@ -49,3 +49,25 @@ it('renders cache statistics', function () {
             (object) ['key' => 'bar', 'hits' => 2, 'misses' => 2],
         ]));
 });
+
+it('does not round numbers up', function () {
+    for ($i = 0; $i < 20_000; $i++) {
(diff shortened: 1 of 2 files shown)


tests/Feature/Livewire/CacheTest.php
```diff
@@ -49,3 +49,16 @@
...
+
+it('does not round numbers up', function () {
+    for ($i = 0; $i < 20_000; $i++) {
+        Pulse::record('cache_hit', 'foo')->count();
+    }
+    Pulse::record('cache_miss', 'foo')->count();
+    Pulse::store();
+
+    Livewire::test(Cache::class, ['lazy' => false])
+        ->dump()
```
jessarcher, line 61:
```suggestion
```

jessarcher:
This will output trailing 0's, e.g. `99.90` - I think we want to remove those?

timacdonald:
Good catch. Fixed.

PR #165, opened 2023-12-06
[1.x] Collect server metrics on Windows

kohenkatz:
This PR adds support for collecting CPU and Memory information on Windows, in addition to the existing Linux and macOS support.

It uses the built-in `wmic` (Windows Management Instrumentation Command-line) utility to collect the information, and the built-in `more` command to remove the header line from the result.

This has been tested on Windows 11, but it should work on all other Windows versions commonly in use.

### DIFF
--- a/src/Recorders/Servers.php
+++ b/src/Recorders/Servers.php
@@ -45,14 +45,14 @@ class Servers
         $memoryTotal = match (PHP_OS_FAMILY) {
             'Darwin' => intval(`sysctl hw.memsize | grep -Eo '[0-9]+'` / 1024 / 1024),
             'Linux' => intval(`cat /proc/meminfo | grep MemTotal | grep -E -o '[0-9]+'` / 1024),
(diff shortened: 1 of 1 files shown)


jbrooksuk (approved):
My pal @sambenne ran the individual commands and the output was as expected. Windows 11.

jessarcher (approved):
Nice! A bit tricky for me to test, but it sounds like it's going to work and is undoubtedly going to work better than not having it at all :+1:

QWp6t:
note that michaelsoft deprecated wmic 2.5 years ago and they expect users to use powershell as an alternative for accessing wmi.

SCR-20231207-jbmu


https://learn.microsoft.com/en-us/windows/win32/wmisdk/wmic

kohenkatz:
> note that michaelsoft deprecated wmic 2.5 years ago and they expect users to use powershell as an alternative for accessing wmi.

They won't be *removing* the utility from Windows 10 or 11 though, they just won't be doing new development on it.

That just means we might need a new way to do this in Windows 12, but I think we can cross that bridge when we come to it.

Issue #167, opened 2023-12-06
SQLSTATE[22001]: String data, right truncated: 1406 Data too long for column 'key' 


mferrara:
### Pulse Version

1.0.0-beta5

### Laravel Version

10.35.0

### PHP Version

8.2

### Livewire Version

3.2.6

### Database Driver & Version

MySQL 8 (Managed Digital Ocean)

### Description

I'm running in an issue with ingesting and the following errors are thrown..

`SQLSTATE[22001]: String data, right truncated: 1406 Data too long for column 'key' at row 159 (Connection: pulse, SQL: insert into `pulse_entries` (`key`, `timestamp`, `type`, `value`) values`

I've removed the rest of the query as there's lots of stuff there I don't want public but it looks like some _very_ long mysql queries are being used as the key and causing the issue. In my application I've replaced those queries and the issue persists, so it seems as though the ingest jobs will continue to retry.

### Steps To Reproduce

Run a MySQL query with a `whereHasIn()` with a few hundred/thousand keys (long enough to be longer than the max length of a TEXT column) in it - the query entry to the `pulse_entries` table should end up failing.

timacdonald:
@mferrara I would recommend publishing the migrations and changing the `TEXT` columns to `MEDIUMTEXT`.

We might also consider changing this as the default column type if we see some more reports of this issue.

Sorry for the troubles you've had there.

rikvdlooi:
Just reporting that we experienced this issue as well.
Thanks for the recommendation Tim.

timacdonald:
No worries. I think we will make these `mediumtext` by default today.

Issue #160, opened 2023-12-06
Dashboard in the browser is very slow


incon:
### Pulse Version

1.0.0-beta5

### Laravel Version

10.35.0

### PHP Version

8.2.13

### Database Driver & Version

MySQL 8.0.33 (DBnign)

### Description

Dashboard is very slow when updating and resizing the window.

https://web.dev/articles/avoid-large-complex-layouts-and-layout-thrashing?utm_source=devtools#avoid-forced-synchronous-layouts

<img width="704" alt="Screenshot 2023-12-06 at 2 26 29 pm" src="https://github.com/laravel/pulse/assets/405629/fd888af7-3f6c-42ca-9c43-d47302fa3bfa">

### Steps To Reproduce

Wait for an update or resize the browser window.

jessarcher:
I'm having trouble replicating this one.

I recorded and let it run for just over 15 seconds to capture a few updates.

![image](https://github.com/laravel/pulse/assets/4977161/3f333920-247a-4187-b481-4f548ebb9188)

If I resize the window constantly while recording then I can get the rendering and painting numbers up, but I assume that's to be expected as I'm forcing it to constantly re-render for the duration of the recording.

incon:
Chrome Version 120.0.6099.62 (Official Build) (arm64)
Apple M2 Max

https://github.com/laravel/pulse/assets/405629/ef4756f3-65da-4dfe-a195-192702a81ef5

timacdonald:
Wonder if this is related to the SQL syntax highlighting.

@incon do you have a bunch of queries in the slow queries table? If you remove some does it improve in speed? If you only have one is it _better_, and if have none is it fast again?

jessarcher:
#172 is released now. Feel free to re-open if there are still issues :+1:

incon:
While much improved with `PULSE_SLOW_QUERIES_HIGHLIGHTING=false` it's still slow.

timacdonald:
@incon do you have a bunch of queries in the slow queries table? If you remove some does it improve in speed? If you only have one is it better, and if have none is it fast again?

incon:
@timacdonald Removing them makes the dashboard fast again.

crynobone:
https://github.com/laravel/pulse/blob/9eda727da25be60c838fb604daced2073c1607fd/config/pulse.php#L191C1-L200C11

@incon did you also update the `config/pulse.php` containing the new options?

incon:
[...]

PR #172, opened 2023-12-07
[1.x] Add the ability to disable SQL highlighting

timacdonald:
Some SQL queries being captured are rather long, as seen in https://github.com/laravel/pulse/issues/167 where the query overflows the TEXT column.

We've also seen some performance issues in the browser, likely due to syntax highlighting: https://github.com/laravel/pulse/issues/160

This adds a flag to disable syntax highlighting of SQL queries on the fly.

### DIFF
--- a/src/Livewire/SlowQueries.php
+++ b/src/Livewire/SlowQueries.php
@@ -56,7 +56,10 @@ class SlowQueries extends Card
         return View::make('pulse::livewire.slow-queries', [
             'time' => $time,
             'runAt' => $runAt,
-            'config' => Config::get('pulse.recorders.'.SlowQueriesRecorder::class),
+            'config' => [
+                'highlighting' => true,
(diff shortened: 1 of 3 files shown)


src/Livewire/SlowQueries.php
```diff
@@ -56,7 +56,10 @@ public function render(): Renderable
         return View::make('pulse::livewire.slow-queries', [
             'time' => $time,
             'runAt' => $runAt,
-            'config' => Config::get('pulse.recorders.'.SlowQueriesRecorder::class),
+            'config' => [
+                'highlighting' => true,
+                ...Config::get('pulse.recorders.'.SlowQueriesRecorder::class),
+            ],
```
timacdonald, line 62: I've done this so that we don't break existing installs. We could remove it once we tag a 1.0 if we tell everyone to re-publish the config.

PR #173, opened 2023-12-07
[1.x] Improve enabled flag

timacdonald:
Currently, if you "disable" Pulse it acts as if you never installed it.

- No events are registered
- The dashboard is not accessible
- The commands are not accessible
- The pulse class is not registered as a singleton
etc.

This might seem like a good idea, but i think it is more problematic.

Imagine you are running Pulse and you are having troubles. You decide to "disable" pulse and redeploy your application.

If you had `php artisan pulse:restart` in your deploy script, you deployment just broke because we don't register that command any more.

I think that the enabled switch should really only impact the recorders. We already document that this is what the flag does.

Screenshot 2023-12-07 at 2 06 36 pm

## Question

One question I do have, though, is if the `pulse:work` command and `pulse:check` command should first look to see if pulse is enabled before starting?

### DIFF
--- a/src/PulseServiceProvider.php
+++ b/src/PulseServiceProvider.php
@@ -36,10 +36,6 @@ class PulseServiceProvider extends ServiceProvider
             __DIR__.'/../config/pulse.php', 'pulse'
         );
 
-        if (! $this->app->make('config')->get('pulse.enabled')) {
-            return;
-        }
-
         $this->app->singleton(Pulse::class);
         $this->app->bind(Storage::class, DatabaseStorage::class);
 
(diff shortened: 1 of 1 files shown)


jessarcher:
> One question I do have, though, is if the pulse:work command and pulse:check command should first look to see if pulse is enabled before starting?

I think `pulse:work` should continue to work regardless of the setting. You might disable recording but still want to process your ingest backlog. If you don't want that to happen, you can stop the command. I.e. it would be more flexible.

Similarly, I think `pulse:check` probably shouldn't fail to start/continue - any supervisor stuff will just keep trying to bring it back. My question is whether it should continue to emit beat events?

Another question is if/how this should affect direct calls to `Pulse::record` that aren't in a recorder? E.g. in the user's controllers/jobs. I think it should probably be a no-op.

timacdonald:
As we chatted about, we are fine with beats being emitted, as the recorders are disabled.

We've also made sure that we disable recording on boot when `PULSE_ENABLED` is `false`.

jessarcher (approved):

PR #174, opened 2023-12-07
[1.]x Allow configurable trimming

timacdonald:
If the `pulse:work` command fails the Ingest will continue to build up in size.

With Redis this is problematic as it is keeping itself in memory. When the `pulse:work` command fails we already trim the ingest on a lottery whenever we put data into the ingest, but it will always retain 7 days worth of data.

That could be problematic for some application depending on their traffic size.

This PR allows the maximum retention for the ingest to be configured.

This value is respected whenever the ingest is trimmed, so it is not guaranteed that the ingest will only contain the configured value until trimming occurs, which again, may happen on a lottery basis.


config/pulse.php
```diff
@@ -83,6 +83,7 @@
         'redis' => [
             'connection' => env('PULSE_REDIS_CONNECTION'),
             'chunk' => 1000,
+            'max_retention' => env('PULSE_REDIS_MAX_RETENTION', '7 day'),
```
timacdonald, line 86: Would love thoughts on this naming.

config/pulse.php
```diff
@@ -83,6 +83,7 @@
         'redis' => [
             'connection' => env('PULSE_REDIS_CONNECTION'),
             'chunk' => 1000,
+            'max_retention' => env('PULSE_REDIS_MAX_RETENTION', '7 days'),
```
timacdonald, line 86: Love thoughts on this naming.

src/Ingests/RedisIngest.php
```diff
@@ -107,4 +108,14 @@ protected function connection(): RedisAdapter
...
+
+    /**
+     * The timestamp, in milliseconds, to trim before.
+     */
+    protected function trimBefore(): int
+    {
+        return CarbonImmutable::now()->subMilliseconds(
+            CarbonInterval::fromString($this->config->get('pulse.ingest.redis.max_retention', '7 days'))->totalMilliseconds
+        )->getTimestampMs();
+    }
```
timacdonald, line 120: I'm continuing to reach for `CarbonImmutable` here because `CarbonInterval` doesn't have any testing helpers. There is a string format hook, but it is deprecated and feels dicey.

config/pulse.php
jessarcher: Do we need the word "max" at all?

timacdonald:
replaced by https://github.com/laravel/pulse/pull/184

Issue #177, opened 2023-12-07
Testing on SQLite with Pulse installed


mralston:
### Pulse Version

v1.0.0-beta5

### Laravel Version

 v10.34.2

### PHP Version

8.2.10

### Livewire Version

v3.2.6

### Database Driver & Version

SQLite 3.41.1 (pdo_sqlite library) on macOS 14.1.1 arm64

### Description

Pulse doesn't support SQLite.

Fine, no problem - I don't wish to run Pulse in SQLite anyway.

However - I run my tests through Pest using an SQLite database.

When Pest runs the database migrations, the create_pulse_tables migration throws the following error:

```
Unsupported database driver [sqlite].

  at database/migrations/2023_06_07_000001_create_pulse_tables.php:34
     30▕             $table->text('key');
     31▕             match ($driver = $connection->getDriverName()) {
     32▕                 'mysql' => $table->char('key_hash', 16)->charset('binary')->virtualAs('unhex(md5(`key`))'),
     33▕                 'pgsql' => $table->uuid('key_hash')->storedAs('md5("key")::uuid'),
  ➜  34▕                 default => throw new RuntimeException("Unsupported database driver [{$driver}]."),
     35▕             };
     36▕             $table->text('value');
     37▕ 
     38▕             $table->index('timestamp'); // For trimming...
```

I would expect the same result through vanilla PHPUnit, but haven't tested that directly.

Is there a better way of doing this? Can the migration check to see whether it's running in the testing environment and fail silently?

### Steps To Reproduce

1. Create a new Laravel project
2. Install Pest
3. Install Laravel Pulse
4. Run `php artisan test`

mralston:
See pull request https://github.com/laravel/pulse/pull/178.

PR #178, opened 2023-12-07
Allow for application testing using SQLite

mralston:
Issue https://github.com/laravel/pulse/issues/177

Laravel Pulse doesn't support SQLite.

It is common to use SQLite whilst running tests of an application built using Laravel.

Pulse's migration throws an error if the database driver is SQLite, causing application tests to fail.

This pull request adds an `App::environment('testing')` check which allows SQLite during testing. Running the migration in any other environment will throw the original exception.

Here is the modified code:

```php
match ($driver = $connection->getDriverName()) {
    'mysql' => $table->char('key_hash', 16)->charset('binary')->virtualAs('unhex(md5(`key`))'),
    'pgsql' => $table->uuid('key_hash')->storedAs('md5("key")::uuid'),
    'sqlite' => App::environment('testing') ? $table->char('key_hash', 16) : throw new RuntimeException("Unsupported database driver [{$driver}]."),
    default => throw new RuntimeException("Unsupported database driver [{$driver}]."),
};
```

### DIFF
--- a/database/migrations/2023_06_07_000001_create_pulse_tables.php
+++ b/database/migrations/2023_06_07_000001_create_pulse_tables.php
@@ -31,7 +31,7 @@ return new class extends Migration
             match ($driver = $connection->getDriverName()) {
                 'mysql' => $table->char('key_hash', 16)->charset('binary')->virtualAs('unhex(md5(`key`))'),
                 'pgsql' => $table->uuid('key_hash')->storedAs('md5("key")::uuid'),
-                default => throw new RuntimeException("Unsupported database driver [{$driver}]."),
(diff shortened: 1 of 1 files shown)


driesvints:
Just disable Pulse during tests with `PULSE_ENABLED`?

jessarcher:
Thanks for the PR!

I'm not opposed to adding SQLite support so Pulse works locally, but I'd want to see it fully implemented (i.e. SQLite-specific expressions for the `upsert*` methods) along with CI tests.

Agree with @driesvints solution if you just want to disable Pulse locally.

mralston:
Thanks for taking the time to look at the PR @jessarcher. Very much appreciated. :)

PR #179, opened 2023-12-07
[1.x] Fix avatar size

sakanjo:
Before | After
:-------------------------:|:-------------------------:
Before  |  After

### PS: This problem seems only to occur when the image dimensions is not equal e.g. 3840x2400.

### DIFF
--- a/resources/views/livewire/usage.blade.php
+++ b/resources/views/livewire/usage.blade.php
@@ -43,7 +43,7 @@
                     <x-pulse::user-card wire:key="{{ $userRequestCount->user->id }}" :name="$userRequestCount->user->name" :extra="$userRequestCount->user->extra">
                         @if ($userRequestCount->user->avatar ?? false)
                             <x-slot:avatar>
-                                <img src="{{ $userRequestCount->user->avatar }}" loading="lazy" class="rounded-full w-8 h-8">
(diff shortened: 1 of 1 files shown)


jbrooksuk (approved):

resources/views/livewire/usage.blade.php
```diff
@@ -43,7 +43,7 @@ class="flex-1"
                     <x-pulse::user-card wire:key="{{ $userRequestCount->user->id }}" :name="$userRequestCount->user->name" :extra="$userRequestCount->user->extra">
                         @if ($userRequestCount->user->avatar ?? false)
                             <x-slot:avatar>
-                                <img height="32" width="32" src="{{ $userRequestCount->user->avatar }}" loading="lazy" class="rounded-full">
+                                <img src="{{ $userRequestCount->user->avatar }}" loading="lazy" class="rounded-full w-8 h-8">
```
jessarcher, line 46: I think it would be good to add `object-cover` so that it keeps the image aspect ratio and crops whatever won't fit in 1:1

```suggestion
                                <img src="{{ $userRequestCount->user->avatar }}" loading="lazy" class="rounded-full w-8 h-8 object-cover">
```

jessarcher (approved):

PR #184, opened 2023-12-08
[1.x] Allow configurable trim length

timacdonald:
If the `pulse:work` command fails the Ingest will continue to build up in size.

With Redis this is problematic as it is keeping itself in memory. When the `pulse:work` command fails we already trim the ingest on a lottery whenever we put data into the ingest, but it will always retain 7 days worth of data.

That could be problematic for some application depending on their traffic size.

This PR allows the maximum retention for the ingest to be configured.

This value is respected whenever the ingest is trimmed, so it is not guaranteed that the ingest will only contain the configured value until trimming occurs, which again, may happen on a lottery basis.

You may specifiy either a time period or a number of entries to retain.

```php
    // ...

    'ingest' => [
        'trim' => [
            // Keep an interval...
            'keep' => '7 days',
            'keep' => '3 days',
            'keep' => '1 day',
            'keep' => '1 hour',

            // Keep a count of entries...
            'keep' => 500_000,
            'lottery' => [1, 1_000],
        ],
    ],
    
    // ...
```

### DIFF
--- a/src/Pulse.php
+++ b/src/Pulse.php
@@ -297,6 +297,7 @@ class Pulse
             $this->entries->filter($this->shouldRecord(...)),
         ));
 
+        // TODO remove fallback when tagging v1
         $odds = $this->app->make('config')->get('pulse.ingest.trim.lottery') ?? $this->app->make('config')->get('pulse.ingest.trim_lottery');
 
         Lottery::odds(...$odds)
--- a/src/Pulse.php
+++ b/src/Pulse.php
@@ -297,7 +297,9 @@ class Pulse
             $this->entries->filter($this->shouldRecord(...)),
         ));
 
-        Lottery::odds(...$this->app->make('config')->get('pulse.ingest.trim_lottery'))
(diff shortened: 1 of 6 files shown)


src/Pulse.php
```diff
@@ -301,7 +301,9 @@ public function store(): int
             $this->entries->filter($this->shouldRecord(...)),
         ));
 
-        Lottery::odds(...$this->app->make('config')->get('pulse.ingest.trim_lottery'))
+        $odds = $this->app->make('config')->get('pulse.ingest.trim.lottery') ?? $this->app->make('config')->get('pulse.ingest.trim_lottery');
+
+        Lottery::odds(...$odds)
```
timacdonald, line 306: Maintaining backwards compatibility for those apps already out there.

We can ditch this when we tag v1. Added to upgrade guide.

timacdonald:
Leaving as a draft, as the upgrade guide will conflict with another PR I have open.

taylorotwell:
@timacdonald some conflicts here

Issue #183, opened 2023-12-07
The Queue card shows SQS queue twice (one with the queued and one with processed and processing items)


BernhardK91:
### Pulse Version

1.0.0-beta5

### Laravel Version

10.34.2

### PHP Version

8.2.13

### Livewire Version

3.2.6

### Description

We are running in an AWS environment and using SQS as queue driver. Currently, we have 1 queue, but Pulse shows it twice. The upper one has the "queued" jobs and the second one has the "processing" and "processed" jobs.

![image](https://github.com/laravel/pulse/assets/5526334/638f894c-02a2-4df5-86d5-6f3ca026c871)

The database (`pulse_aggregates`) shows that the SQS entries are logged with different names. Some with the driver name and some with the full SQS address:

![image](https://github.com/laravel/pulse/assets/5526334/3c6d00db-02a9-46cd-8da0-296711c6fdde)

### Steps To Reproduce

Currently, I do not have a second project that uses SQS queue, so I cannot share reproduction steps.

timacdonald:
We are aware of this and will get it sorted in the next beta. Thanks for reporting.

timacdonald:
This will be fixed in https://github.com/laravel/pulse/pull/187

We can follow that PR.

PR #187, opened 2023-12-08
[1.x] Normalize SQS queue names

timacdonald:
Fixes https://github.com/laravel/pulse/issues/183

### DIFF
--- a/src/Recorders/Queues.php
+++ b/src/Recorders/Queues.php
@@ -9,6 +9,7 @@ use Illuminate\Queue\Events\JobProcessed;
 use Illuminate\Queue\Events\JobProcessing;
 use Illuminate\Queue\Events\JobQueued;
 use Illuminate\Queue\Events\JobReleasedAfterException;
+use Illuminate\Support\Str;
 use Laravel\Pulse\Pulse;
 
 /**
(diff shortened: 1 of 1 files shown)


src/Recorders/Queues.php
```diff
@@ -97,4 +105,30 @@ protected function getDefaultQueue(string $connection): string
...
+        }
+
+        if ($config['suffix'] ?? null) {
+            $suffix = preg_quote($config['suffix'], '#');
+
+            $queue = preg_replace("#{$suffix}$#", '', $queue) ?? $queue;
+        }
+
+        return $queue;
+    }
```
timacdonald, line 133: Made this method naming SQS specific. If we have to do more normalizing we can introduce a "normalizeQueue" method that wraps other methods.

The `preg_replace(...) ?? $queue` null coalesce thingo is there because `preg_replace` can fail and return `null`.

src/Recorders/Queues.php
timacdonald, line 1: I've moved the extra work, such as `getDefaultQueue`, into the `lazy` call to postpone it to the last minute.

jbrooksuk (approved):
Tested on Vapor 👍🏻

PR #188, opened 2023-12-08
[1.x] Add "sum" and "min" aggregations

jessarcher:
This PR adds support for "sum" and "min" aggregations, in addition to the "max", "avg", and "count" aggregations we already support.

Also refactored some of the logic to not loop over the entries more than needed, and abstracted the repeated preaggregation logic now that we have five methods.

Cc @aarondfrancis


aarondfrancis:
This is gonna be super helpful. Thank you so much!

PR #205, opened 2023-12-11
[1.x] Add notification to Usage card when unable to resolve users

jessarcher:
This PR introduces a helpful notification when users are unable to be resolved for the `usage` card.

Previous improvements have made it so that the lack of a `name` or `email` field on the model does not cause an error and will instead display the user's name as "Unknown". This notification is intended to complement that by directing them to the appropriate section of the documentation.

|Dark mode|Light mode|
|-|-|
|image|image|


jessarcher:
Just holding off on this one until we figure out some potential improvements to the user recording/resolving story with different user providers, multi-tenancy, etc.

jessarcher:
I am going to close this for now as #251 has solved a lot more scenarios and changed stuff that this PR was depending on. If we get more reports of folks having issues we can revisit this notification.

PR #211, opened 2023-12-11
add commands to collect server metrics from BSD systems

sergyu91ma:
add commands to collect server metrics from BSD systems


taylorotwell:
We are exploring just allowing a custom callback for this for full customization.

sergyu91ma:
I see. Then, I'll wait. Thanks.

PR #212, opened 2023-12-11
[1.x] Bail on store when entries is empty

timacdonald:
If there are no entries to write, we will bail early

This means we won't try and trim the database when not entries are being written - but I think this is okay - good even.

### DIFF
--- a/src/Pulse.php
+++ b/src/Pulse.php
@@ -129,18 +129,16 @@ class Pulse
             ->filter(fn ($recorder) => $recorder->listen ?? null)
             ->each(fn ($recorder) => $event->listen(
                 $recorder->listen,
-                fn ($event) => $this->rescue(fn () => Collection::wrap($recorder->record($event)))
+                fn ($event) => $this->rescue(fn () => $recorder->record($event))
             ))
         );
 
         $recorders
             ->filter(fn ($recorder) => method_exists($recorder, 'register'))
             ->each(function ($recorder) {
-                $record = function (...$args) use ($recorder) {
-                    $this->rescue(fn () => Collection::wrap($recorder->record(...$args)));
-                };
-
-                $this->app->call($recorder->register(...), ['record' => $record]);
+                $this->app->call($recorder->register(...), [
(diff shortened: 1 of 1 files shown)


src/Pulse.php
```diff
@@ -129,18 +129,16 @@ public function register(array $recorders): self
             ->filter(fn ($recorder) => $recorder->listen ?? null)
             ->each(fn ($recorder) => $event->listen(
                 $recorder->listen,
-                fn ($event) => $this->rescue(fn () => Collection::wrap($recorder->record($event)))
+                fn ($event) => $this->rescue(fn () => $recorder->record($event))
```
timacdonald, line 132: `Collection::wrap` is a remnant from when recorders used to return entries. They no longer do.

src/Pulse.php
```diff
@@ -129,18 +129,16 @@ public function register(array $recorders): self
...
             ->filter(fn ($recorder) => method_exists($recorder, 'register'))
             ->each(function ($recorder) {
-                $record = function (...$args) use ($recorder) {
-                    $this->rescue(fn () => Collection::wrap($recorder->record(...$args)));
-                };
-
-                $this->app->call($recorder->register(...), ['record' => $record]);
+                $this->app->call($recorder->register(...), [
+                    'record' => fn (...$args) => $this->rescue(fn () => $recorder->record(...$args)),
+                ]);
```
timacdonald, line 141: 🧹 `Collection::wrap` is a remnant from when recorders used to return entries. They no longer do. 

Inlined as it felt nice.

src/Pulse.php
```diff
@@ -508,15 +521,20 @@ public function handleExceptionsUsing(callable $callback): self
...
+    public function rescue(callable $callback): mixed
     {
         try {
-            $callback();
+            return $callback();
         } catch (Throwable $e) {
             ($this->handleExceptionsUsing ?? fn () => null)($e);
         }
+
+        return null;
```
timacdonald, line 537: I've checked all places where this could be impactful.

The only place it could be an issue is if someone returned `false` from a recorder. It could bail an `$collection->each()` call, but "don't hold it that way"

PR #217, opened 2023-12-12
Amend link in upgrade guide

ashleyshenton:
I was checking the upgrade guide to see what will need actioning when 1.x is released and noticed the first link under required pointed to the wrong PR.


jessarcher:
Thanks!

PR #224, opened 2023-12-13
[1.x] Adds option to run check and work commands once

joedixon:


src/Commands/WorkCommand.php
```diff
@@ -57,7 +59,7 @@ public function handle(
                 $lastTrimmedStorageAt = $now;
             }
 
-            if ($this->option('stop-when-empty')) {
+            if ($this->option('stop-when-empty') || $this->option('once')) {
```
timacdonald, line 62: I don't think we need this. `stop-when-empty` serves this purpose.

I think `once` is misleading in this context, as it might imply to only work one record, similar to `queue:work --once` flags to only run one job.

That is why we went for `stop-when-empty`.

src/Commands/CheckCommand.php
```diff
@@ -61,18 +68,33 @@ public function handle(
...
+     * Check the current server's pulse.
+     */
+    protected function check(
+        Pulse $pulse,
+        CacheStoreResolver $cache,
+        Dispatcher $event,
+        ?CarbonImmutable $now = null,
+        ?CarbonInterval $interval = null
+    ): void {
+        $now ?: CarbonImmutable::now();
```
crynobone, line 85:
```suggestion
        $now ??= CarbonImmutable::now();
```

src/Commands/WorkCommand.php
joedixon: Yeah, I was on the fence with that too, but the goal here is to be able to run these commands from the scheduler and it felt consistent to me at the time to be able to schedule them like so:

```php
$schedule->command('pulse:work --once')->everySecond();
$schedule->command('pulse:check --once')->everyFiveSeconds();
```

Having slept on it and seen your comment, I think you're right and it's more likely to be misleading so I'll remove it.

timacdonald:
Closing in favour of https://github.com/laravel/pulse/pull/314/

PR #226, opened 2023-12-14
Fixes PHPStan error

crynobone:
See https://github.com/laravel/pulse/actions/runs/7198402317/job/19607755653


src/Pulse.php
```diff
@@ -493,6 +493,8 @@ public function js(): string
 
     /**
      * The default "vendor" cache keys that should be ignored by Pulse.
+     *
+     * @return array<int, string>
```
jessarcher, line 497:
```suggestion
     * @return list<string>
```

PR #229, opened 2023-12-14
[1.x] Use `PULSE_PATH` env instead of fixed text for `/pulse`

rozsazoltan:
In the configuration file, we statically refer to the `/pulse` path during restrictions. However, **the path should be dynamically determined from the `PULSE_PATH` environment variable**.

### DIFF
--- a/config/pulse.php
+++ b/config/pulse.php
@@ -204,7 +204,7 @@ return [
             'sample_rate' => env('PULSE_SLOW_REQUESTS_SAMPLE_RATE', 1),
             'threshold' => env('PULSE_SLOW_REQUESTS_THRESHOLD', 1000),
             'ignore' => [
-                '#^/' . env('PULSE_PATH', 'pulse') . '$#', // Pulse dashboard...
+                '#^/'.env('PULSE_PATH', 'pulse').'$#', // Pulse dashboard...
             ],
         ],
 
--- a/config/pulse.php
+++ b/config/pulse.php
@@ -204,7 +204,7 @@ return [
             'sample_rate' => env('PULSE_SLOW_REQUESTS_SAMPLE_RATE', 1),
             'threshold' => env('PULSE_SLOW_REQUESTS_THRESHOLD', 1000),
             'ignore' => [
-                '#^/pulse$#', // Pulse dashboard...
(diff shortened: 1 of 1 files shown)


jbrooksuk (approved):

config/pulse.php
```diff
@@ -204,7 +204,7 @@
             'sample_rate' => env('PULSE_SLOW_REQUESTS_SAMPLE_RATE', 1),
             'threshold' => env('PULSE_SLOW_REQUESTS_THRESHOLD', 1000),
             'ignore' => [
-                '#^/pulse$#', // Pulse dashboard...
+                '#^/' . env('PULSE_PATH', 'pulse') . '$#', // Pulse dashboard...
```
siarheipashkevich, line 207: @rozsazoltan what do you think about extracting `'pulse'` string to the constant and use this constant in all places where needed default path? Because you found good point where we omit all places where PULSE_PATH is used

siarheipashkevich: @rozsazoltan and what if I declared `PULSE_PATH=/pulse-custom` in my env config, this regexp will work correctly ? because we will have to `//`

rozsazoltan: > @rozsazoltan what do you think about extracting `'pulse'` string to the constant and use this constant in all places where needed default path? Because you found good point where we omit all places where PULSE_PATH is used

I was thinking about this as well, in the end, for the sake of a shorter code modification, I didn't highlight the `pulse` path, but we can do that if desired.

siarheipashkevich: I agree with you, but we need the opinion of the core team members here.

taylorotwell:
Mark as ready for review when it's ready.

rozsazoltan:
Thank you, okay! Once the https://github.com/laravel/pulse/pull/231 is approved, this is also ready for merge.

Note to myself: If the mentioned PR cannot be implemented, at least a `trim()` will be necessary (see https://github.com/laravel/pulse/pull/229#discussion_r1426674310).

timacdonald (commented):
I'm not sure we need this in the default config, tbh.

If you have a requirement to change the default path you can make the change in your own application's config.

rozsazoltan:
Related: #343

PR #233, opened 2023-12-14
[1.x] Improve migration errors to guide developers with unsupported drivers

timacdonald:
### `phpunit` + `sqlite` + `PULSE_ENABLED=false`

Database migrations are ignored ✅

Screenshot 2023-12-15 at 9 59 04 am

---

### `phpunit` + `sqlite` + `PULSE_ENABLED=true`

Throws an exception pointing developer to disable pulse if they want to.

Screenshot 2023-12-15 at 9 50 29 am

---

### `artisan migrate` + `sqlite` + `PULSE_ENABLED=true`

Throws an exception indicating that the DB is not supported.

Screenshot 2023-12-15 at 9 57 52 am

---

### `artisan migration` + `sqlite` + `PULSE_ENABLED=false`

Throws an exception indicating that the DB is not supported. This is important because we don't want to record in the migrations table that the migration has been run, because enabling pulse later will not work.

Screenshot 2023-12-15 at 9 58 27 am

### DIFF
--- a/database/migrations/2023_06_07_000001_create_pulse_tables.php
+++ b/database/migrations/2023_06_07_000001_create_pulse_tables.php
@@ -1,12 +1,10 @@
 <?php
 
-use Illuminate\Database\Migrations\Migration;
 use Illuminate\Database\Schema\Blueprint;
-use Illuminate\Support\Facades\Config;
-use Illuminate\Support\Facades\DB;
 use Illuminate\Support\Facades\Schema;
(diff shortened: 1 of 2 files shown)


timacdonald:
I think we should revert https://github.com/laravel/laravel/commit/4705136d1b90c708097d57f0959802772d65cb69 after this is merged, but again, not a hill I'm gonna die on.

PR #234, opened 2023-12-14
[1.x] Ignore all internal pulse activity

timacdonald:
This PR addresses memory leaks where Pulse's internal activity creates new entries that are never persisted. It also addresses memory leaks present in the `pulse:work` command.

### DIFF
--- a/src/Pulse.php
+++ b/src/Pulse.php
@@ -597,8 +597,6 @@ class Pulse
      */
     public function __call($method, $parameters): mixed
     {
-        $storage = $this->app->make(Storage::class);
-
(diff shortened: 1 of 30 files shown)


src/Pulse.php
```diff
@@ -286,13 +286,14 @@ public function filter(callable $filter): self
...
-     * Store the queued items.
+     * Ingest the entries.
      */
-    public function store(): int
+    public function ingest(): int
     {
         $entries = $this->rescue(function () {
             $this->lazy->each(fn ($lazy) => $lazy());
 
+            // here as well.
```
timacdonald, line 296: Need a failing test here.

jessarcher (approved):

PR #235, opened 2023-12-15
[1.x] Add SQLite support for local environments

jessarcher:
This PR introduces SQLite support for Pulse.

**This is intended for local development only,** primarily to prevent any errors when using SQLite for local development or testing, but also to allow developers to have Pulse working locally when setting up authorization gates, custom cards, etc.

Note that SQLite does not appear to include any hashing functions out of the box, so the hash columns must be generated in PHP.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -85,8 +85,9 @@ jobs:
     strategy:
       fail-fast: true
       matrix:
-        php: [8.2, 8.3]
+        php: [8.3]
         laravel: [10]
+        stability: [prefer-stable]
 
     name: PHP ${{ matrix.php }} - Laravel ${{ matrix.laravel }} - Stability ${{ matrix.stability }} - PostgreSQL 14
 
@@ -130,9 +131,9 @@ jobs:
     strategy:
       fail-fast: true
       matrix:
-        php: [8.1, 8.2, 8.3]
+        php: [8.3]
         laravel: [10]
-        stability: [prefer-lowest, prefer-stable]
+        stability: [prefer-stable]
 
     name: PHP ${{ matrix.php }} - Laravel ${{ matrix.laravel }} - Stability ${{ matrix.stability }} - SQLite
 
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -116,3 +116,47 @@ jobs:
         env:
           DB_CONNECTION: pgsql
           DB_PASSWORD: password
+
+  sqlite:
+    runs-on: ubuntu-22.04
+
+    services:
+      redis:
+        image: redis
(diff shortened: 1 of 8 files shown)


jessarcher:
Leaving this as a draft until #234 is merged as it will have some conflicts.

.github/workflows/tests.yml
```diff
@@ -85,8 +85,9 @@ jobs:
     strategy:
       fail-fast: true
       matrix:
-        php: [8.2, 8.3]
+        php: [8.3]
```
jessarcher, line 88: The `mysql` job matrix covers all of the supported PHP versions so there didn't seem to be much pointing in running all of the versions against Postgres and SQLite.

database/migrations/2023_06_07_000001_create_pulse_tables.php
```diff
@@ -23,6 +23,7 @@ public function up(): void
             match ($this->driver()) {
                 'mysql' => $table->char('key_hash', 16)->charset('binary')->virtualAs('unhex(md5(`key`))'),
                 'pgsql' => $table->uuid('key_hash')->storedAs('md5("key")::uuid'),
+                'sqlite' => $table->string('key_hash'),
```
jessarcher, line 26: SQLite doesn't have many data types so I think a varchar is as good as it gets. It also doesn't have any hashing functions built-in, so the hashing needs to be in PHP.

timacdonald (approved):

bakerkretzmar:
> This is intended for local development only

Any specific reasons/gotchas to avoid using this in non-local environments? If an app is on one server and small enough to use SQLite without running into issues around writes conflicting, would it be fine with Pulse too?

jessarcher:
> > This is intended for local development only
> 
> Any specific reasons/gotchas to avoid using this in non-local environments? If an app is on one server and small enough to use SQLite without running into issues around writes conflicting, would it be fine with Pulse too?

Honestly, I am not sure. Conflicts were the main issue I ran into when using SQLite on a multi-user app, so Pulse would likely increase the chances of that happening as it could be inserting/updating on most requests.

I included the warning because I only had local environments in mind when building the feature, and I haven't tested it in a production scenario. It might work great, but I don't want to make any assurances about it. I'd be keen to hear how it works out if you try it. If there are any issues with a simple fix, feel free to send a PR too.

Wulfheart:
@jessarcher is there a way contributions could make it ready for production?

nunomaduro:
For info, I am using Pulse and SQLite in production.

bakerkretzmar:
Same, been using it with SQLite for several months in production and it's going great.

Wulfheart:
I think then it should have a little hint in the docs.

PR #245, opened 2023-12-19
[1.x] Null ingest

timacdonald:
Because Pulse is active in production, it is also good to have it enabled in your testsuite.

This introduces a "null" driver that may be used in testsuites to keep the logic parts of Pulse active, but have the persistence side of things ignored to improve performance of your local testsuite.

### DIFF
--- a/src/PulseServiceProvider.php
+++ b/src/PulseServiceProvider.php
@@ -18,6 +18,7 @@ use Illuminate\View\Compilers\BladeCompiler;
 use Illuminate\View\Factory as ViewFactory;
 use Laravel\Pulse\Contracts\Ingest;
 use Laravel\Pulse\Contracts\Storage;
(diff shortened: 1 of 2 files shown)


src/PulseServiceProvider.php
```diff
@@ -52,6 +53,7 @@ protected function registerIngest(): void
         $this->app->bind(Ingest::class, fn (Application $app) => match ($app->make('config')->get('pulse.ingest.driver')) {
             'storage' => $app->make(StorageIngest::class),
             'redis' => $app->make(RedisIngest::class),
+            null, 'null' => $app->make(NullIngest::class),
```
timacdonald, line 56: phpunit.xml and .env will provide us with `null`, while our first party configs often hardcode the string `'null'` to indicate the null driver should be used, as seen in `config/logging.php` and `config/broadcasting.php`.

Issue #200, opened 2023-12-09
Running test with mysql database throws too many connections or memory issue


walterdis:
### Pulse Version

latest

### Laravel Version

10.32

### PHP Version

8.1.9

### Livewire Version

dont know

### Database Driver & Version

MySQL 8.0.31 for macOS 14.0 on arm64 (Docker)

### Description

Hello,

When I run tests using the MySQL database with --parallel, even with Redis, the tests stop in the middle, giving me the error message: "Too many connections." 
```
[1040] Too many connections (Connection: mysql, SQL: drop database if exists `project_testing_test_2`))
```

If i run without --parallel, it's giving me the memory limit exception
```
PHP Fatal error:  Allowed memory size of 134217728 bytes exhausted (tried to allocate 20480 bytes) in /var/www/vendor/laravel/framework/src/Illuminate/Collections/Arr.php on line 565

Fatal error: Allowed memory size of 134217728 bytes exhausted (tried to allocate 20480 bytes) in /var/www/vendor/laravel/framework/src/Illuminate/Collections/Arr.php on line 565
PHP Fatal error:  Allowed memory size of 134217728 bytes exhausted (tried to allocate 32768 bytes) in /var/www/vendor/pestphp/pest/src/Plugins/Actions/CallsShutdown.php on line 1

Fatal error: Allowed memory size of 134217728 bytes exhausted (tried to allocate 32768 bytes) in /var/www/vendor/pestphp/pest/src/Plugins/Actions/CallsShutdown.php on line 1
```

Am I doing something wrong?

### Steps To Reproduce

Configure tests to use mysql
- php artisan test --parallel (too many connections)
- php artisan test (throws memory exception)

jbrooksuk:
Disable Pulse in your tests. You can add `PULSE_ENABLED=false` in your `phpunit.xml` file.

walterdis:
Hello.

I've already done that and the problem still persists

timacdonald:
@walterdis can you run the following terminal command in your application and let me know the output.

```
composer outdated -- laravel/pulse | grep version
```

jessarcher:
Have you configured a different database connection for Pulse but maybe pointing to the same MySQL server? It's the only reason I can think why Pulse's existence would increase the number of database connections.

walterdis:
> @walterdis can you run the following terminal command in your application and let me know the output.
> 
> ```
> composer outdated -- laravel/pulse | grep beta
> ```

@timacdonald  Hello ;)

[...]

PR #250, opened 2023-12-20
[1.x] Introduce buffer size

timacdonald:
fixes https://github.com/laravel/pulse/issues/200

We have built Pulse to always optimize for faster performance over lower memory usage.

One example of this is that we capture entry / value objects (or "lazy" closures) during the request and postpone ingesting them until the request has been sent the client.

For most cases this is an ideal tradeoff. A little more memory usage for faster performance. However for some testsuite work and also some production cases this approach introduces other issues where memory limits may be exceeded because the sheer number of objects captured continues to grow.

Take a feature test that does `70,000` loops performing a DB insert in each iteration. With Pulse installed and active this loop now also captures `70,000` objects, including the queries SQL, etc.

The options for fixing this are:

1. Setting `PULSE_ENABLED=false` in their testsuite

This is okay, but because Pulse runs in production it makes sense that it should also be able to run in your testsuite. Additionally, this case, however rare, may also happen in production where we don't want to disable Pulse.

2. "Ignore" the work done.

```diff
+ Pulse::ignore(function () {

    foreach (range(0, 70_000) as $item) {
        DB::table('foo')->insert([...]);
    }

+ });
```

This is okay, but you basically have to hit a memory overflow in production to know where to ignore.

This PR addresses the issue for testsuites and production applications by introducing a buffer size. Whenever a new entry or closure is captured we determine if the pending number of entries is over the buffer size - if it is we will ingest all pending records to free up memory. This allows applications to tune how they want the tradeoff to work.

Using too much memory? You'll need to make your buffer size smaller. Ingesting too often and have memory capacity left? Increase your buffer size.

For most applications you won't even need to care about this. A single request will only be capturing a handful of entries. I would imagine `100` entries to be a very high number of entries in MOST of cases and non of this will every be a concern.

However, testsuite and some more edge-case production endpoints will need to consider the buffer size.

## A note of filtering

We offer the ability for users to filter entries before they are ingested.

```php
Pulse::filter(function (Entry|Value $entry): bool {
    // ...
});
```

To improve performance during the request we postpone running the filters until we Ingest. Previously this was only done at the end of Laravel's lifecycle, e.g., after the response has been sent to the client.

Now that it is possible for an Ingest to occur during a request, i.e., when the buffer size overflows, the filtering may happen during a request.

The tradeoff we have here is that when an ingest is triggered because the buffer has been exceeded - after the filtering is done on 5,000 entries in the buffer there may only be 5 entries to Ingest.

I still feel this is the best approach to improve performance for the 95% use-case - but I wanted to mention it here.


config/pulse.php
```diff
@@ -79,6 +79,8 @@
     'ingest' => [
         'driver' => env('PULSE_INGEST_DRIVER', 'storage'),
 
+        'buffer' => env('PULSE_INGEST_BUFFER', 5_000),
+
```
timacdonald, line 83: This is totally arbitrary. I don't know what a good "default" for this would even be.

`5_000` felt good at the time 🤷‍♂️

src/Storage/DatabaseStorage.php
```diff
@@ -351,6 +352,17 @@ protected function preaggregateAverages(Collection $entries): Collection
...
+     * Collapse values.
+     *
+     * @param  \Illuminate\Support\Collection<int, \Laravel\Pulse\Value>  $values
+     * @return \Illuminate\Support\Collection<int, \Laravel\Pulse\Value>
+     */
+    protected function collapseValues(Collection $values): Collection
+    {
+        return $values->reverse()->unique(fn (Value $value) => [$value->key, $value->type]);
+    }
+
```
timacdonald, line 365: This was added because one of my tests failed in Postgres 14.

See https://github.com/laravel/pulse/pull/250/files#diff-82b9572841700c12fb8fdbbc144371ae2af52f68529f5b0cbccc1acc3910fb10R428-R446

jessarcher (approved):

PR #251, opened 2023-12-20
[1.x] Allow complete control of user resolution

### LINKED ISSUE #121: Having multiple Authenticatable models is incorrectly tracked
### Pulse Version

v1.0.0-beta2

### Laravel Version

10.26.2

### PHP Version

8.2.13

### Database Driver & Version

_No response_

### Description

User requests are logged with only the ID value in the key-field, however it is possible to have multiple Authenticatable models in Laravel.

The default behavior for Pulse seems to be to log all Authenticated requests, also for other Authenticatable models, but not differentiate the model type.

This leads to a bug where request count are aggregated towards the User model with the ID even if another model is making the authenticated request.

### Steps To Reproduce

Install Laravel Pulse on a project with a custom Authenticatable model that can login for instance using Laravel\Sanctum\HasApiTokens

jessarcher:
This PR introduces a new approach for resolving users in Pulse.

There are a few scenarios we're aiming to solve here:

#### 1. User models that don't have a `name` or `email` field, or that want to provide an alternative avatar

The most commonly needed Pulse customization is to change what fields are used to display users in the dashboard. The current approach requires you to resolve the users yourself and be sure to include an `id` field so we can map them back up again:

```php
Pulse::users(fn ($ids) => User::findMany($ids)->map(fn ($user) => [
    'id' => $user->id,
    'name' => $user->name,
    'extra' => $user->email,
    'avatar' => $user->avatar_url,
]));
```

The proposed `Pulse::user` method allows you to hook into just the field customization instead:

```php
Pulse::user(fn (Authenticatable $user) => [
    'name' => $user->name,
    'extra' => $user->email,
    'avatar' => $user->avatar_url,
]);
```

We have maintained the existing API for backward compatibility but might propose removing it for the stable release.

#### 2. Authenticatable classes other than `App\Models\User` or `App\User`

Pulse currently checks for an `App\Models\User` or `App\User` model. With this PR, Pulse will use the configured user provider (from `config/auth.php`) to resolve the user. This means we'll support whichever model has been configured and will also support other user providers, such as the `DatabaseUserProvider`.

#### 3. Complex authentication where the authenticable ID alone is not unique or enough to resolve the user (e.g. multi-tenancy)

Pulse currently only captures the authenticated user ID when attributing records to a user. For some applications this is not enough to uniquely resolve back to a user record for display in the dashboard.

This PR introduces a new `ResolvesUsers` contract which you can implement and bind in the service container to completely control what user identifier is captured, and how that identifier is resolved back to a user for the dashboard.

The contract defines three methods required to capture and resolve users:

```php
use Illuminate\Contracts\Auth\Authenticatable;
use Illuminate\Support\Collection;
use Laravel\Pulse\Contracts\ResolvesUsers;

class CustomResolver implements ResolvesUsers
{
    public function key(Authenticatable $user): int|string|null
    {
        return json_encode([$user->tenantId, $user->id]);
    }

    public function load(Collection $keys): self
    {
        // Optionally eager load the given keys...

        return $this;
    }

    public function fields(int|string|null $key): array
    {
        // Locate the user and return the required fields...

        return [
            'name' => '...',
            'extra' => '...',
            'avatar' => '...',
        ];
    }
});
```

Fixes #121


src/Pulse.php
```diff
@@ -361,37 +348,41 @@ public function recorders(): Collection
...
+        return $resolver->load($keys);
+    }
 
-        return $ids->map(fn (string|int $id) => [
-            'id' => $id,
-            'name' => "User ID: {$id}",
-        ]);
+    /**
+     * Resolve the users' details using the given closure.
+     *
```
timacdonald, line 362: Maybe?

```suggestion
     * @deprecated
```

timacdonald (approved):

src/PulseServiceProvider.php
```diff
@@ -40,6 +41,7 @@ public function register(): void
 
         $this->app->singleton(Pulse::class);
         $this->app->bind(Storage::class, DatabaseStorage::class);
+        $this->app->singletonIf(ResolvesUsers::class, Users::class);
```
timacdonald, line 44: Far out, Laravel. Love this.

PR #264, opened 2024-01-04
[1.x] Add mariadb test workflow

AustinW:
This PR simply adds support for MariaDB. Not many changes as most of the configurations are congruent with MySQL.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -62,6 +62,59 @@ jobs:
         env:
           DB_CONNECTION: mysql
           DB_USERNAME: root
+  
+  mariadb:
+    runs-on: ubuntu-22.04
+
+    services:
+      mysql:
+        image: mariadb:10.5
+        env:
+          MYSQL_ALLOW_EMPTY_PASSWORD: yes
+          MYSQL_DATABASE: forge
+        ports:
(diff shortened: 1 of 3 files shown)


timacdonald:
@AustinW Pulse already supports MariaDB via the `mysql` driver.

I've remove the configuration changes and removed the matrix from the MariaDB workflow.

We currently only run the full matrix against MySQL and then run once against the other DB types.

timacdonald (approved):

AustinW:
@timacdonald We use this package in our project: https://github.com/ybr-nx/laravel-mariadb?tab=readme-ov-file#configure-only-larvel-53-and-54 which requires us to set the driver to `mariadb`. We had to modify the migration manually for the migration to run. It's not a big deal, but it'd be nice to include that case in the migration.

timacdonald:
I see. As that is not a first party package I don't think we should account for it in our base migration. Will ask the team for their thoughts.

jessarcher:
> I see. As that is not a first party package I don't think we should account for it in our base migration. Will ask the team for their thoughts.

I agree. It's easy enough to tweak the migration if needed. That package is pretty old, too - I wonder whether it's needed any longer?

Issue #263, opened 2024-01-04
Is it possible to publish the assets?


sunscreem:
My Nginx config automatically caches and optimises assets, such as `livewire/livewire.js?id=e2b302e9`.

This means a route which fakes the assets being available there, such as the one pulse uses, returns a 404.

In other projects I've been able to run `php artisan livewire:publish` to work around this, but in Pulse this does nothing. Would this be an option going forward?

jbrooksuk:
This is a dynamic route that is registered by Livewire and serves the `livewire.js` script. You may want to update your Nginx script to ignore this script from being cached / optimized.

driesvints:
Thanks @jbrooksuk

sinnbeck:
There is guide here: https://benjamincrozat.com/livewire-js-404-not-found

sunscreem:
Thanks @jbrooksuk and @sinnbeck  - it's set by my hosting company so I was looking for a workaround to save having to go to them to start changing things. Thanks anyway.

sinnbeck:
@sunscreem Just ran the publish method and it does not seem to work in livewire 3. So maybe ask on the livewire repo or make a pr to add it back?

timacdonald:
Opened a PR to address this moving forward: https://github.com/laravel/pulse/pull/266

Issue #201, opened 2023-12-09
PAge is not loading missing livewire.js


nezaboravi:
### Pulse Version

1

### Laravel Version

10.35.0

### PHP Version

8.2

### Livewire Version

3

### Database Driver & Version

mysql

### Description

ON production site, with empty database entries, it does loading page.

IN dev console i see that livewire.js gives 404.
Now, this is weird, because the same code, works on dev server and on my local.

Now, maybe is worth to mention, i am running producion site via Forge as isolated site? Dont know if this can have any affects.

### Steps To Reproduce

NOthinf special here to reproduce it. As in my bug description. Happy to provide access to repo if needed.

haugstrup:
I had the same issue and it was an nginx configuration issue. Go to edit your site's nginx configuration file and find the `# Cache expires headers` section near the bottom.

Immediately **before** that section insert this snippet to tell nginx to let your Laravel app handle requests for `livewire.js`

```
    location = /livewire/livewire.js {
        expires off;
        try_files $uri $uri/ /index.php?$query_string;
    }
```

nezaboravi:
Thank you @haugstrup for reply and advise, but i am not confident to do that, as this is Laravel Forge provided server, same as my dev server, both forge provided, and works on dev, but not on live. Also works on local. The only difference is that Live server is created with isolation flag.

jbrooksuk:
Please provide us with the exact version of Pulse you're running. v1.0.0-beta5 is currently the latest version, so check you're on that too.

nezaboravi:
@jbrooksuk here
<img width="443" alt="image" src="https://github.com/laravel/pulse/assets/1515172/7cb2d954-4d91-4bf9-a1e9-1c5e144cac2a">

jessarcher:
@nezaboravi Could you please provide the output of:

```sh
php artisan route:list --path=livewire
```

> Happy to provide access to repo if needed.

The repo would be helpful if you're able!

timacdonald:
This is a known "issue" with Livewire.

See their docs...

<img width="771" alt="Screenshot 2023-12-11 at 11 13 42 am" src="https://github.com/laravel/pulse/assets/24803032/f9011643-60a7-4706-9bd0-5896089e15ed">

which links to: https://benjamincrozat.com/livewire-js-404-not-found

We are considering bundling Livewire into Pulse so that users do not need to consider or deal with this. For now, follow the above blogpost guide on configuring NGINX.

[...]

Issue #164, opened 2023-12-06
net::ERR_ABORTED 404 (Not Found)


sandesh556:
### Pulse Version

1.0.0-beta-5

### Laravel Version

10.35.0

### PHP Version

8.2.0

### Description

![Screen Shot 2023-12-06 at 3 14 41 PM](https://github.com/laravel/pulse/assets/79452740/b08d953e-aafc-4be4-951e-ad10cd5d7fb4)
![Screen Shot 2023-12-06 at 3 18 08 PM](https://github.com/laravel/pulse/assets/79452740/badd6775-526a-40e9-9e91-1aee19542ec8)
![Screen Shot 2023-12-06 at 3 29 40 PM](https://github.com/laravel/pulse/assets/79452740/4dc9b8d5-62f3-4780-8bab-4989d5aa7e41)

Also saw some existing issues created for this, none of those helped, also noticed none of them had a path in the url

### Steps To Reproduce

So installed the new version of Pulse and did all the necessary steps to publish it. And then when I get to the dashboard this is what I see. But the APP_URL has a path associated with it and the URL called in the console doesn't take that path Pulse

maloun96:
I have that issue too.

sinnbeck:
Check that the route actually exists with `php artisan route:list`. It is a route that livewire adds.

It should look like this
```
GET|HEAD        livewire/livewire.js ........................................................................................................................................................... Livewire\Mechanisms › FrontendAssets@returnJavaScriptAsFil

sinnbeck:
Just noticed that you have `/public` in your url. Seems you didnt set up your webserver to point to the public directory

![image](https://github.com/laravel/pulse/assets/13980140/5fe1c797-49f3-48fb-81e8-ef9b10f52ce4)

Fix this first.

maloun96:
https://admin.blablabla/livewire/livewire.js?id=f477dd12. I can open in a new tab, I see the code, but fetching with XHR results 404.
I alos have the route

sandesh556:
> @sandesh556 Just noticed that you have `/public` in your url. Seems you didnt set up your webserver to point to the public directory
> 
[...]

Issue #145, opened 2023-12-05
Pulse 1.0.0-beta3 version is not working due to appending APP_URL.


ahaneef29:
### Pulse Version

1.0.0-beta3

### Laravel Version

10.34.2 

### PHP Version

8.2.13 

### Database Driver & Version

Mysql

### Description

I installed Pulse 1.0.0-beta3 version today which not loading Pulse dashboard due to the appending APP_URL., if i commented the APP_URL from .env file. Its working fine. i tried production server too and its same.

Please check the screenshot.
<img width="1753" alt="pulse screenshot" src="https://github.com/laravel/pulse/assets/10325179/a6c45531-8a9f-465b-8a8c-50aacab95ac9">

### Steps To Reproduce

Install the Pulse 1.0.0-beta3 version into existing Laravel app.

jbrooksuk:
I'm not able to replicate this myself. Also, the Request URL doesn't look like it's related to Pulse, as it's simply your `APP_URL` and a cache query of some kind.

Can you try disabling Debugbar too?

sinnbeck:
Do you perhaps have a /config/livewire.php file? Check if it has a `asset_url` set. If it does, try removing it.. It will overwrite livewire's own `/livewire/livewire.js` url

ahaneef29:
@sinnbeck Yes, I have updated `asset_url` to `null` in /config/livewire.php file

`'asset_url' => null, `

Which is working fine.

Thank you @sinnbeck @jbrooksuk

sinnbeck:
@ahaneef29 Just remove it completely. Its not part of the livewire config file by default

Check the default here
https://github.com/livewire/livewire/blob/main/config/livewire.php

ahaneef29:
@sinnbeck removed Thank you again.

jessarcher:
Is this resolved now? If so, please close the issue. Thanks!

driesvints:
Guess this is resolved, thanks.

timacdonald:
Opened a PR to address this moving forward: https://github.com/laravel/pulse/pull/266

PR #266, opened 2024-01-05
[1.x] Manually inline Livewire assets

### LINKED ISSUE #85: livewire.js gives 404 in production
### Pulse Version

v1.0.0-beta2

### Laravel Version

10.34.2

### PHP Version

8.2.10

### Description

In production, the asset `/livewire/livewire.js?id=eaa5c323` in the Network tab shows up as 404 Not Found. If I paste the full  URL (including the domain name) to the asset in the browser, it does load. This does not happen locally.

Because of this, the dashboard only shows the skeleton loading state.

### Steps To Reproduce

Deploy to production I guess 😅

timacdonald:
We've had numerous reports of issues loading Livewire's dynamic `/livewire/livewire.js` route on deployed environments.

fixes #85, #131, #145, #164, #201, #263

This is a known issue that Livewire documents.

Screenshot 2024-01-05 at 2 54 10 pm

We have been telling people to update their NGINX configuration, however it would be nice if Pulse "just worked".

Livewire's recommend solution is to build their scripts into your `app.js` via a build tool, such as Vite. We cannot do this as we might build against Livewire `3.0.1` while the end user may install Livewire `3.3.0` that requires different built assets.

Instead we can inline Livewire's assets via `file_get_contents`. We will always output the correct Livewire assets for the currently installed Livewire version.

Having the assets cached at a URL is not really any benefit to Pulse. It makes sense for a full-blown Livewire application, but for a single page dashboard like Pulse it is fine to inline the assets and download them on every dashboard visit as you would usually be viewing the same page for some time and only occasionally doing full page reloads. We also do not need to consider SEO performance.


package-lock.json
timacdonald, line 1: I initially thought we should remove our lock file, but we need it for our build step that commits the built assets.

I've updated the dependencies while I was looking at all this.

resources/views/components/pulse.blade.php
```diff
@@ -50,7 +47,7 @@
             </main>
         </div>
 
-        @livewireScripts
-        @stack('scripts')
```
timacdonald, line 54: The `@stack('scripts')` was a remnant from our previous iterations. We no longer use this.

resources/views/components/pulse.blade.php
```diff
@@ -50,7 +47,7 @@
...
+        {!! Laravel\Pulse\Facades\Pulse::js() !!}
+        @livewireScriptConfig
```
timacdonald, line 51: > When Livewire detects the @livewireScriptConfig directive, it will refrain from injecting the Livewire and Alpine scripts. If you are using the @livewireScripts directive to manually load Livewire, be sure to remove it. Make sure to add the @livewireStyles directive if it is not already present.

resources/views/components/pulse.blade.php
```diff
@@ -13,10 +13,10 @@
...
 
         {!! Laravel\Pulse\Facades\Pulse::css() !!}
-
         @livewireStyles
 
         {!! Laravel\Pulse\Facades\Pulse::js() !!}
+        @livewireScriptConfig
```
timacdonald, line 19: > When Livewire detects the @livewireScriptConfig directive, it will refrain from injecting the Livewire and Alpine scripts. If you are using the @livewireScripts directive to manually load Livewire, be sure to remove it. Make sure to add the @livewireStyles directive if it is not already present.

src/Pulse.php
```diff
@@ -520,11 +520,17 @@ public function css(string|Htmlable|array|null $css = null): string|self
...
-        if (($content = file_get_contents(__DIR__.'/../dist/pulse.js')) === false) {
+        if (
+            ($livewire = @file_get_contents(__DIR__.'/../../../livewire/livewire/dist/livewire.js')) === false &&
+            ($livewire = @file_get_contents(__DIR__.'/../vendor/livewire/livewire/dist/livewire.js')) === false) {
+            throw new RuntimeException('Unable to load the Livewire JavaScript.');
+        }
+
+        if (($pulse = @file_get_contents(__DIR__.'/../dist/pulse.js')) === false) {
             throw new RuntimeException('Unable to load the Pulse dashboard JavaScript.');
         }
```
timacdonald, line 531: This exception would never be thrown without the `@`

src/Pulse.php
```diff
@@ -520,11 +520,17 @@ public function css(string|Htmlable|array|null $css = null): string|self
      */
     public function js(): string
     {
-        if (($content = file_get_contents(__DIR__.'/../dist/pulse.js')) === false) {
+        if (
+            ($livewire = @file_get_contents(__DIR__.'/../../../livewire/livewire/dist/livewire.js')) === false &&
+            ($livewire = @file_get_contents(__DIR__.'/../vendor/livewire/livewire/dist/livewire.js')) === false) {
+            throw new RuntimeException('Unable to load the Livewire JavaScript.');
+        }
```
timacdonald, line 527: These two are required to support applications with Pulse installed and local testing.

jessarcher: Would it be better to use `base_path()`?

timacdonald: I thought that as well, however TestBench changes the `base_path` to its own directory. This fix is specifically for those scenarios so it doesn't work here unfortunately.

package-lock.json
timacdonald, line 1: I initially thought we should remove our lock file, but we need it for our build step that commits the built assets.

I've updated the dependencies while I was looking at all this.

jessarcher (approved):

jbrooksuk (approved):

PR #267, opened 2024-01-05
Use (potentially overloaded) key() method for finding user

hulkur:
This way it is not nessessary to change `find()` method for custom user resolver.


jessarcher (approved):
Thanks! Makes sense to me!

PR #271, opened 2024-01-10
[1.x] Reduce chances of missing Pulse records due to deadlocks when not using Redis

jessarcher:
It's possible (but rare) for Pulse to get a deadlock if two processes upsert the same data concurrently when there is no existing row to update.

Pulse catches the exception, so the application is not impacted. However, it will result in Pulse missing records.

It's already possible to solve the issue by using the Redis ingest, as it removes concurrent upserts, but this PR aims to reduce the chances for those not using Redis. It does this by:

1. Increasing the number of attempts on the transaction to 3 so that if it does occur, there's a good chance the data will still be captured.
2. Moving data processing (like pre-aggregation) outside of the transaction to reduce time spent in the transaction.

We replicated the issue locally by running multiple processes, all capturing the same data repeatedly in a loop. Prior to the change, we could reproduce the issue every 2-3 minutes using 8 concurrent processes. After increasing the attempts, we didn't see the issue in the 30 minutes we left it running.

### DIFF
--- a/src/Storage/DatabaseStorage.php
+++ b/src/Storage/DatabaseStorage.php
@@ -45,76 +45,77 @@ class DatabaseStorage implements Storage
             return;
         }
 
-        $this->connection()->transaction(function () use ($items) {
-            [$entries, $values] = $items->partition(fn (Entry|Value $entry) => $entry instanceof Entry);
+        [$entries, $values] = $items->partition(fn (Entry|Value $entry) => $entry instanceof Entry);
 
-            $entries
-                ->reject->isOnlyBuckets()
-                ->chunk($this->config->get('pulse.storage.database.chunk'))
-                ->each(fn ($chunk) => $this->connection()
-                    ->table('pulse_entries')
-                    ->insert(
-                        $this->requiresManualKeyHash()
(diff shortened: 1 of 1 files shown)


src/Storage/DatabaseStorage.php
```diff
@@ -45,76 +45,77 @@ public function store(Collection $items): void
...
+            ->chunk($this->config->get('pulse.storage.database.chunk'));
+
+        $minimumChunks = $this->preaggregateMinimums(collect($minimums)) // @phpstan-ignore argument.templateType argument.templateType
+            ->chunk($this->config->get('pulse.storage.database.chunk'));
+
+        $maximumChunks = $this->preaggregateMaximums(collect($maximums)) // @phpstan-ignore argument.templateType argument.templateType
+            ->chunk($this->config->get('pulse.storage.database.chunk'));
+
+        $sumChunks = $this->preaggregateSums(collect($sums)) // @phpstan-ignore argument.templateType argument.templateType
+            ->chunk($this->config->get('pulse.storage.database.chunk'));
```
timacdonald, line 82: "sum chunks" made me chuckle.

timacdonald (approved):

PR #274, opened 2024-01-12
Add BSD support for Servers Recorder

paulo-hortelan:
This PR introduces BSD support to record server information:
- CPU usage
- Memory usage

I tested on a real BSD environment and there was some divergencies about how to get the real used memory information. Even though [Zabbix documentation](https://www.zabbix.com/documentation/current/en/manual/appendix/items/vm.memory.size_params) says that `used memory = active + wired + cached` It's not what I have seen on my tests. 

It looks like `active + inactive + cached`  has more accurate values about the used memory, especially when It gets increased.

**PS:** I'm open to fix this in case someone shows me wrong.

### DIFF
--- a/src/Recorders/Servers.php
+++ b/src/Recorders/Servers.php
@@ -46,6 +46,7 @@ class Servers
             'Darwin' => intval(`sysctl hw.memsize | grep -Eo '[0-9]+'` / 1024 / 1024),
             'Linux' => intval(`cat /proc/meminfo | grep MemTotal | grep -E -o '[0-9]+'` / 1024),
             'Windows' => intval(((int) trim(`wmic ComputerSystem get TotalPhysicalMemory | more +1`)) / 1024 / 1024),
+            'BSD' => intval(`sysctl hw.physmem | grep -Eo '[0-9]+'` / 1024 / 1024),
             default => throw new RuntimeException('The pulse:check command does not currently support '.PHP_OS_FAMILY),
         };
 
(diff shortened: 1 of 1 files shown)


taylorotwell:
Drafting for @jessarcher review.

jessarcher:
Hey @paulo-hortelan,

I don't have a handy BSD system to test this, but I assume it's okay.

I am curious whether you've tried the Darwin commands with BSD? I'm not sure how much Darwin has diverged from standard BSD, but it would be nice to use the same for both if it makes sense.

paulo-hortelan:
I've tried, and they are very similar. The commands are just a little different, for example:

- On `sysctl hw.memsize`, the `hw.memsize` doesn't exist on BSD, instead we have `physmem`, `usermem`, `realmem`
- The Darwin command `vm_stat | grep 'Pages free'` became `vmstat -s | grep 'pages free'`
- The Darwin command `top -l 1 | grep -E "^CPU"` became `top -b -d 2| grep 'CPU:'`

But the main difference is the way the memory is calculated. I'm not sure about Darwin but on BSD just doing `totalMemory - pagesFree*pageSize` doesn't capture the real memory usage because It keeps changing the allocated memory between states (inactive, laundry, active, free, etc.). 

When I tried to copy the Darwin system to calculate memory on my BSD system, which has 8gb, It was showing around 7gb usage when the real usage (calculated by Zabbix) was around 700mb.

jessarcher (approved):

PR #285, opened 2024-01-18
[1.x] Remove unused variable declaration

alirezadp10:
As `registersRoutes` is a boolean variable, I guess it would be better to change ‍‍‍`registersRoutes` to `isRoutesRegistered` as it conveys the purpose more accurately.

### DIFF
--- a/src/PulseServiceProvider.php
+++ b/src/PulseServiceProvider.php
@@ -66,7 +66,7 @@ class PulseServiceProvider extends ServiceProvider
      */
     public function boot(): void
     {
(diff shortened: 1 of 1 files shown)


timacdonald:
I think it is fine.

We use `registersRoutes` across our other packages in the ecosystem: https://github.com/search?q=org%3Alaravel%20registersRoutes&type=code

This is also an internal facing method. There isn't really an reason user-land apps would call this.

alirezadp10:
Yes, you're right. I dropped that commit and just kept the last one.

timacdonald (approved):
We can ignore the static analysis error. That has been fixed in `1.x`

PR #299, opened 2024-01-29
Carbon v3

driesvints:

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -31,8 +31,8 @@ jobs:
     strategy:
       fail-fast: true
       matrix:
-        php: [8.1, 8.2, 8.3]
-        laravel: [10]
+        php: [8.2, 8.3]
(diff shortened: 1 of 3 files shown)


nunomaduro (approved):

timacdonald:
Pushed a fix to maintain the `int` value returned from `diffInMilliseconds`. This return type changed in Carbon v3.

Although not a major issue, it is a potential change in functionality if the config were to contain a float value.

timacdonald (approved):

timacdonald:
I would prefer to keep this in draft until v3 is actually released and we know for sure what all the breaking changes are.

Otherwise we merge this and they tag additional breaking changes in future beta release and we have to re-review.

driesvints:
@timacdonald Should only be bug fixes from now on normally.

driesvints:
Stable release out now.

Issue #315, opened 2024-02-14
Option to ignore offline servers


leventdev:
When deploying an app that uses Pulse on Kubernetes, every time there's a new deployment, the host name changes and shows as a new server.
This quickly clutters up the pulse site, and even after enough time passes to not have any server data from the selected period (1H, 6H, 24H, 7D), the server is still visible as offline.
It would be neat to have a configuration option for ignoring offline servers all together, or at least after there's no more data from them from the selected time period.
Example screenshot:
![image](https://github.com/laravel/pulse/assets/64219099/74446b69-13b7-452e-9b20-b8c871b72574)

driesvints:
I think we'd appreciate a PR for this, thanks!

DellanX:
Something to add to this, for future developers planning to add Pulse to their Kubernetes cluster, is to set the PULSE_SERVER_NAME to be the node name, which you can do by adding the following snippet to your container spec:
```yaml
          env:
            - name: PULSE_SERVER_NAME
              valueFrom:
                fieldRef:
                  fieldPath: spec.nodeName
```

Furthermore, if you use a Daemon Set instead of a deployment, you can configure there to be exactly one pulse service per node.

This is my configuration:

```yaml
apiVersion: apps/v1
kind: DaemonSet
metadata:
  name: {{ .Release.Name }}-pulse-check-default
  labels:
    tier: backend
    layer: pulse-check
spec:
  replicas: 1
  selector:
    matchLabels:
      tier: backend
      layer: pulse-check
  template:
    metadata:
      labels:
        tier: backend
        layer: pulse-check
    spec:
      imagePullSecrets:
        - name: {{ .Release.Name }}-registry-key
      containers:
        - name: pulse-check
          image: <docker-registry>:{{ .Values.version }}
          command:
            - php
          args:
            - artisan
            - pulse:check
          ports:
            - containerPort: 9000
          env:
            - name: PULSE_SERVER_NAME
              valueFrom:
                fieldRef:
                  fieldPath: spec.nodeName
          envFrom:
            - configMapRef:
                name: {{ .Release.Name }}
            - secretRef:
                name: {{ .Release.Name }}

```

This will prevent a new server entry from being created each time you update your code 🙂

razvan-p:
Do you have any docker / docker-compose solution?
The PULSE_SERVER_NAME from my docker-compose.yml seems to have string value "{{.Node.Hostname}}", not the host name.
```
services:
    app:
        environment:
[...]

PR #316, opened 2024-02-15
[1.x] Ignore offline servers

leventdev:
## This PR adds the option to ignore offline servers.
Based on #315 

The PR adds a config option, to the Servers recorder, and changes the Servers Livewire card, to remove offline servers if the config option is enabled.

Added options:
PULSE_SERVER_IGNORE_OFFLINE - when true, offline servers will be hidden
PULSE_SERVER_IGNORE_OFFLINE_AFTER - when specified, offline servers will be hidden after the specified time (in minutes). Just specifying this value, will not hide offline servers, unless the other value is set to true.

Tested on a multi server k8s setup, and seems to work as expected.

### Issue description
When deploying an app that uses Pulse on Kubernetes, after every deployment, the Pulse page will be cluttered by a bunch of offline servers.
These servers remain visible, even after the selected timeframe (1H,6H...) does not have any data from the server.
Example screenshot:
image

If the config option is enabled, the servers page will show only online servers:
image

### DIFF
--- a/src/Livewire/Servers.php
+++ b/src/Livewire/Servers.php
@@ -7,6 +7,8 @@ use Illuminate\Contracts\Support\Renderable;
 use Illuminate\Support\Facades\View;
 use Livewire\Attributes\Lazy;
 use Livewire\Livewire;
+use Illuminate\Support\Facades\Config;
+use Laravel\Pulse\Recorders\Servers as ServersRecorder;
 
 /**
  * @internal
@@ -25,6 +27,10 @@ class Servers extends Card
             return $this->values('system')
                 ->map(function ($system, $slug) use ($graphs) {
                     $values = json_decode($system->value, flags: JSON_THROW_ON_ERROR);
+                    if (!CarbonImmutable::createFromTimestamp($system->timestamp)->isAfter(now()->subSeconds(30)) && Config::get('pulse.recorders.' . ServersRecorder::class . '.ignore_offline')) {
+                        //instead of returning null, we remove the value from the list
+                        return null;
+                    }
 
                     return (object) [
                         'name' => (string) $values->name,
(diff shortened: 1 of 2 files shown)


driesvints:
@leventdev it's always best to add more info about this to the PR description instead of linking to an issue. Please be thorough about what this fixes so the people reviewing this have a clear picture.

src/Livewire/Servers.php
```diff
@@ -25,6 +27,10 @@ public function render(): Renderable
             return $this->values('system')
                 ->map(function ($system, $slug) use ($graphs) {
                     $values = json_decode($system->value, flags: JSON_THROW_ON_ERROR);
+                    if (!CarbonImmutable::createFromTimestamp($system->timestamp)->isAfter(now()->subSeconds(30)) && Config::get('pulse.recorders.' . ServersRecorder::class . '.ignore_offline', false)) {
```
jbrooksuk: I think the "offline" period should be configurable.

leventdev: Great idea. Added the option for that.

taylorotwell:
Drafting for review from @timacdonald or @jessarcher

jessarcher:
Dig this! At the moment, the server will appear for roughly 7 days (i.e. until the data is trimmed).

Wondering whether we can reduce it to a single config option, though. E.g. just `ignore_after` with a default value of `null` which would just mean that it's never ignored (until the data is trimmed).

leventdev:
> Dig this! At the moment, the server will appear for roughly 7 days (i.e. until the data is trimmed).
> 
> Wondering whether we can reduce it to a single config option, though. E.g. just `ignore_after` with a default value of `null` which would just mean that it's never ignored (until the data is trimmed).

Using just one config option makes sense, and should be an easy update too.

jessarcher:
Hey @leventdev, just checking in to see if you're planning on making the above change?

leventdev:
Hi @jessarcher, thanks for the reminder!
Yes, I plan on making the changes in (depending on my availability) the coming days.

jessarcher:
Closing in favour of #355 which contains these commits and the requested change :+1:

PR #326, opened 2024-03-06
limit sql length to avoid breaking the json

kima92:
### Issue
We've identified a bug within Laravel Pulse where if the slow query detector captures an excessively large SQL statement, it leads to a disruption in the JSON schema stored in the database. This disruption makes it impossible to read the affected data, impacting the tool's ability to provide accurate insights into application performance.

### Root Cause
The root cause of this issue lies in the default length of the database column designated for storing SQL statements. When a captured SQL statement exceeds this default length (65535), the resulting JSON encoding process creates a payload that violates the column's size constraints, leading to corruption of the JSON schema and subsequent data retrieval problems.

### Proposed Solution
To address this issue, I propose implementing a check to truncate SQL statements to a maximum character length before encoding them into JSON. This approach ensures that the payload size remains within the database field size constraints, thus preserving the integrity of the JSON schema.

### Changes Made
Start using `Str::limit($sql, 65000)` to make sure we are not getting too close to the limit.

### Impact
This fix ensures that Laravel Pulse can handle large SQL statements without compromising the JSON schema's integrity in the database, thereby maintaining the tool's efficacy in providing performance insights.

### DIFF
--- a/config/pulse.php
+++ b/config/pulse.php
@@ -196,7 +196,7 @@ return [
             'threshold' => env('PULSE_SLOW_QUERIES_THRESHOLD', 1000),
             'location' => env('PULSE_SLOW_QUERIES_LOCATION', true),
             'highlighting' => env('PULSE_SLOW_QUERIES_HIGHLIGHTING', true),
-            'max_sql_length' => env('PULSE_SLOW_QUERIES_MAX_SQL_LENGTH', null),
+            'max_query_length' => env('PULSE_SLOW_QUERIES_MAX_QUERY_LENGTH', null),
             'ignore' => [
                 '/(["`])pulse_[\w]+?\1/', // Pulse tables...
                 '/(["`])telescope_[\w]+?\1/', // Telescope tables...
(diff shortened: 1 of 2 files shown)


taylorotwell:
Drafting pending review from @jessarcher or @timacdonald

timacdonald:
@kima92, it sounds like you have the old migration where the `key` was a `TEXT` column. We have migrated it to a `MEDIUMTEXT`, which should mean you can store ~16,777,215 characters, instead of the ~65,535 characters you are hitting up against.

I would recommend updating your schema to match the latest migrations.

That being said, I do feel it would be a nice additional to allow the maximum query length to be controlled. In most cases I would think that there isn't much value in having Pulse retrieve _potentially_ megabytes worth of SQL string.

I would prefer this be a configurable option rather than a hardcoded length.

kima92:
Thank you @timacdonald, I've updated the schema and it works great now.
Also added config with env to the code

PR #329, opened 2024-03-13
[1.x] Add SQL Server support

jessarcher:


jessarcher:
Giving up on this one for now. Getting an error with the upsert query, but I'm thinking it might be an issue with the `key_hash` column from the migration, which I just copied verbatim from the MySQL implementation as a starting point.

PR #330, opened 2024-03-13
[1.x] Add `mariadb` driver support plus Laravel 11 CI tests

jessarcher:
GitHub actions had started failing, which appears to be because we were relying on default `DB_` env vars and/or default config file vars which have changed.

This PR:
* Makes the DB auth settings explicit in CI rather than relying on defaults.
* Adds Laravel 11 to the matrix
* Adds support for the `mariadb` database driver, which, in addition to being nice to have, also solved an issue with different default collations.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -32,7 +32,7 @@ jobs:
       fail-fast: true
       matrix:
         php: [8.1, 8.2, 8.3]
-        laravel: [10]
+        laravel: [10, 11]
         stability: [prefer-lowest, prefer-stable]
 
     name: PHP ${{ matrix.php }} - Laravel ${{ matrix.laravel }} - Stability ${{ matrix.stability }} - MySQL 5.7
@@ -85,7 +85,7 @@ jobs:
       fail-fast: true
       matrix:
         php: [8.3]
-        laravel: [10]
+        laravel: [11]
         stability: [prefer-stable]
 
     name: PHP ${{ matrix.php }} - Laravel ${{ matrix.laravel }} - Stability ${{ matrix.stability }} - MariaDB 10.5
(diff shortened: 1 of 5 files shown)


timacdonald (approved):

.github/workflows/tests.yml
```diff
@@ -61,17 +63,22 @@ jobs:
         run: vendor/bin/pest
         env:
           DB_CONNECTION: mysql
-          DB_USERNAME: root
+          DB_DATABASE: pulse
+          DB_USERNAME: pulse
+          DB_PASSWORD: password
+          DB_COLLATION: utf8mb4_unicode_ci
```
jessarcher, line 69: Setting the credentials explicitly instead of relying on the defaults which recently changed.

The collation needs to be explicitly included here because the default collation is not supported on MySQL 5.7 where these tests are running.

.github/workflows/tests.yml
```diff
@@ -61,17 +63,22 @@ jobs:
...
 
   mariadb:
     runs-on: ubuntu-22.04
 
     services:
       mysql:
-        image: mariadb:10.5
+        image: mariadb:10
```
jessarcher, line 76: It seems MariaDB 10.5 doesn't support the new default collation. Rather than customize the collation here, I've opted to match the framework's CI tests and specify `10`, which would give us MariaDB 10.6.

driesvints (approved):
Good stuff 👍

francoism90:
> Pulse's first-party storage implementation currently requires a MySQL or PostgreSQL database.

Is this still the case after this PR?

driesvints:
@francoism90 no MariaDB was added in here. I'll PR that to the docs.

PR #337, opened 2024-03-15
[1.x] Freeze time to fix flakey tests

timacdonald:
The queue tests can sometimes take a little while, esp. because we spin up a queue worker.

This means that sometimes the current "bucket" would change within the test and cause failures in CI, i.e., we have flakey tests.

This fixes that by freezing time across the board for the queues' tests. We may want to roll this our further, but for now it fixes what we know is causing us issues.


tests/Feature/Recorders/QueuesTest.php
```diff
@@ -376,6 +382,7 @@ function queueAggregates()
     Pulse::ignore(fn () => expect(Queue::size())->toBe(1));
     $aggregates = queueAggregates();
     expect($aggregates)->toHaveCount(12);
+    // Fail...
```
timacdonald:
```suggestion
```

jessarcher (approved):

PR #339, opened 2024-03-16
[1.x] Customise server stats

### LINKED ISSUE #322: Servers inside container
### Pulse Version

1.0^@beta

### Laravel Version

10.39

### PHP Version

8.3

### Livewire Version

3.0.2

### Database Driver & Version

Postgresql 15

### Description

I'm integrating laravel pulse in my docker enviroment in production, and I setted all container to run the command ```php artisan pulse:check ``` in background, but as result in my dashboard is showing like this.

image

All the containers are running in the same machine, so all of then show the same information. I'm thinking if has a way to show only the CPU and Memory of that container? Even if i has to pass a flag to the command like ```php artisan pulse:check --container" this will work for me.

I have this setup because I run the workers in separeted containers, so i can scale then in my application. 

### Steps To Reproduce

Run this command at every container of the aplication in the same machine.

timacdonald:
fixes https://github.com/laravel/pulse/issues/322

Allows the server stat collection to be customised.

In a service provider it is now easy to customise the CPU and memory stat collection depending on your special operating system or unique infrastructure (docker, et al.).

```php
Servers::detectCpuUsing(function (): int {
    // ...
});

Servers::detectMemoryUsing(fn () => [
    'total' => (int) /* ... */,
    'used' => (int) /* ... */,
]);
```

### DIFF
--- a/src/Recorders/Servers.php
+++ b/src/Recorders/Servers.php
@@ -59,7 +59,7 @@ class Servers
     /**
      * Detect memory via the given callback.
      *
-     * @param  null|(callable(): array{total: int, used: int})   $callback
+     * @param  null|(callable(): array{total: int, used: int})  $callback
      */
     public static function detectMemoryUsing(?callable $callback)
     {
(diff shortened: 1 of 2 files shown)


jessarcher:
Dig this!

jessarcher (approved):

justinkekeocha:
any documentation on how this can be used?

PR #340, opened 2024-03-17
[1.x] Allow thresholds of slow queries, jobs and requests to be customised by regex pattern

matheus-carvalho:
I'm facing a problem with slow jobs and requests I've already know is slower than others, so it would be useful I could customise the thresholds individually, using regex patterns.

I think it would be useful to anyone who have resources known to be slow, like report jobs, huge-data endpoints or something else, and want to monitor them with a higher threshold, not just ignore them.

I guess this changes will not break any existing feature because we still with a 'default' threshold, which will be used until some individual customisation.


matheus-carvalho:
Many checks are broken but I can't understand the reason, even with a fresh clone of `1.x` branch it was breaking locally

jessarcher:
Hi @matheus-carvalho,

Thanks for the PR!

I've fixed the failing test. The Livewire payload in the test was incomplete as it was previously inconsequential, but with these changes, the payload needs to be parsed to determine the path for the threshold.

I think this could be a useful feature, but at the moment, it seems like a breaking change for existing users. To move forward and consider this feature, it would need to support the existing config file format where the threshold is a single value rather than an associate array. I'm also not sure whether this new format should be the default in the config file or just something we document.

matheus-carvalho:
Hi @jessarcher,

Thank you for fixing the tests!

About the problem with the config file, what do you think about creating a previous check? We could test if the value is a array to continue this way, in case of negative we could fallback to the current way, just using the single value.

For this change I would need to change the two methods `threshold` in both Concerns `HasThreshold` and `Thresholds`, and a little adjustment at `usage.blade`.

If you think it's viable I could submit the changes.

jessarcher:
@matheus-carvalho Sounds reasonable.

Our other concern is introducing a concept that ties cards to individual recorders with the `recorder` method. Cards can leverage data from multiple recorders (e.g., the `Usage` card), so I think the `threshold` method might need to accept the recorder class or threshold config as an argument.

matheus-carvalho:
Thank you for your attention @jessarcher.

Alright, this would really be a problem, I'll implement the changes to fix these points and submit the commit.

Please let me know if there are something else worrying you.

jessarcher:
Thanks @matheus-carvalho!

I haven't run it locally yet to check out the UI additions, but the only other thing I'm wondering about is whether the "Top 10 users experiencing slow endpoints" view in the Usage card should have something to indicate when there are path-specific thresholds in play. I'm imagining a scenario where a user complains about something being "slow", but it doesn't show up on that card even though it's over the default threshold because it was under the path-specific threshold. The person viewing the dashboard may not be aware of the additional configuration.

It's also up to Taylor whether we want to take this feature on, but it looks useful to me!

matheus-carvalho:
Hello @jessarcher,

I have pushed some improvements to take care of the pointed concerns.
- `threshold` methods now accepts a recorder argument.
- check if the threshold config is an array to maintain compatibility.
- add a information icon in the actions area of `usage` card to tell about customised thresholds.

Here are some images from my local tests so you can take a look at UI additions.

[Information icon](https://github.com/laravel/pulse/assets/41878558/f4f4f0ab-576f-43e9-9a70-13553d6c42f3)
[Slow requests](https://github.com/laravel/pulse/assets/41878558/1c375e79-738b-497b-95a9-0389a63267aa)
[Slow outgoing requests](https://github.com/laravel/pulse/assets/41878558/e834ef62-38db-4af8-8a33-eebc91e06640)
[Slow jobs](https://github.com/laravel/pulse/assets/41878558/bb6ffa0b-1979-446e-a24c-dbcbce89d32c)
[Slow queries](https://github.com/laravel/pulse/assets/41878558/3a245b85-c216-4683-b942-355e8ab452f9)

matheus-carvalho:
Hello,

excuse me but I just want to know if there are something else I could do for this PR.

timacdonald:
We'll review this one shortly and see if we can move it along.

src/Recorders/Concerns/LivewireRoutes.php
```diff
@@ -20,7 +20,7 @@ protected function resolveRoutePath(Request $request): array
         $path = $route->getDomain().Str::start($route->uri(), '/');
         $via = $route->getActionName();
 
-        if ($route->named('*livewire.update')) {
+        if ($route->named('*livewire.update') && $request->has('components.0.snapshot')) {
```
timacdonald: Although I'm not against this change, could you let me know why you made it here?

Did you hit an issue?

src/Recorders/Concerns/Thresholds.php
```diff
@@ -2,13 +2,36 @@
...
+        if (! is_array($config)) {
+            return $config;
+        }
+
+        // @phpstan-ignore argument.templateType, argument.templateType
+        $custom = collect($config)
+            ->except(['default'])
+            ->first(fn ($threshold, $pattern) => preg_match($pattern, $value));
+
+        return $custom ?? $config['default'] ?? 1_000;
```
timacdonald: Notice that we have a hard-coded `1_000ms` here as the default threshold when no default is specified; this matches our config defaults.

When a user changes from a single threshold to an array, if they forget the `default` key that the app continues to work with this default.

Not sure if we wanna keep this or just document that they need to have the `default` key, but this felt okay to me.

src/Recorders/Concerns/LivewireRoutes.php
matheus-carvalho: Hello, this changes comes to fix the problem commented [here](https://github.com/laravel/pulse/pull/340#issuecomment-2002682634). Was authored by @jessarcher [here](https://github.com/laravel/pulse/pull/340/commits/09ed4fb33c5bba560a41dec8bc7ac41c212b130e)

src/Recorders/Concerns/Thresholds.php
matheus-carvalho: I'm not sure if you want my opinion about this, but don't breaking the app even without a `default` key sounds good for me. Although, if you want help documenting the `default` key's mandatory, count on me.

src/Recorders/Concerns/LivewireRoutes.php
timacdonald: Gotcha! Thanks.

resources/views/livewire/cache.blade.php
```diff
@@ -20,7 +20,7 @@
                     Str::plural('group', $count)
                 );
             @endphp
-            <button title="{{ $message }}" @click="alert('{{ str_replace("\n", '\n', $message) }}')">
+            <button title="{{ $message }}" @click="alert(@js($message))">
```
timacdonald, line 23: Standardising use of the `@js` directive, which I've used in some of the changes in this PR.

timacdonald:
Some UI shots here.

I've decided to keep this pretty simple and only change the UI based on the threshold config type. Array = per item threshold reporting. Integer = Global type threshold.

We could have done fancy stuff checking if there was more than one threshold value for the all the listed items - but I didn't like the idea that the UI would change if an entry came into the table that had a different threshold while you were watching it.

## Slow requests

```php
'threshold' => 1000,
```

Screenshot 2024-05-09 at 10 46 38

```php
'threshold' => [
    'default' => 1000,
    '#pattern#' => 10,
],
```

Screenshot 2024-05-09 at 10 46 18

## Application usage card

```php
'threshold' => 1000,
```

Screenshot 2024-05-09 at 10 46 46


```php
'threshold' => [
    'default' => 1000,
    '#pattern#' => 10,
],
```

Screenshot 2024-05-09 at 10 47 05

Screenshot 2024-05-09 at 10 47 11

## Slow queries

```php
'threshold' => 1000,
```

Screenshot 2024-05-09 at 10 50 18


```php
'threshold' => [
    'default' => 1000,
    '#pattern#' => 10,
],
```

Screenshot 2024-05-09 at 10 49 59

## Outgoing requests


```php
'threshold' => 1000,
```
Screenshot 2024-05-09 at 10 52 30

```php
'threshold' => [
    'default' => 1000,
    '#pattern#' => 10,
],
```

Screenshot 2024-05-09 at 10 53 06

## Slow Jobs
```php
'threshold' => 1000,
```

Screenshot 2024-05-09 at 10 55 15


```php
'threshold' => [
    'default' => 1000,
    '#pattern#' => 10,
],
```

Screenshot 2024-05-09 at 10 55 48

timacdonald (approved):
Fan of this feature. Thanks!

PR #344, opened 2024-03-25
[1.x] Adding $keys parameter to graph()

cord:
Allow to select and limit which key-series to include into graph() data.

### DIFF
--- a/src/Storage/DatabaseStorage.php
+++ b/src/Storage/DatabaseStorage.php
@@ -447,7 +447,7 @@ class DatabaseStorage implements Storage
      *
      * @param  list<string>  $types
      * @param  'count'|'min'|'max'|'sum'|'avg'  $aggregate
-     * @param  list<string>  $keys  
+     * @param  list<string>|null  $keys  
      * @return \Illuminate\Support\Collection<string, \Illuminate\Support\Collection<string, \Illuminate\Support\Collection<string, int|null>>>
      */
     public function graph(array $types, string $aggregate, CarbonInterval $interval, ?array $keys = null): Collection
@@ -469,10 +469,10 @@ class DatabaseStorage implements Storage
 
         $structure = collect($types)->mapWithKeys(fn ($type) => [$type => $padding]);
 
-        return $this->connection()->table('pulse_aggregates') // @phpstan-ignore return.type
+        return $this->connection()->table('pulse_aggregates')
             ->select(['bucket', 'type', 'key', 'value'])
             ->whereIn('type', $types)
-            ->when($keys, fn($query) => $query->whereIn('key_hash', collect($keys)->map(fn($key) => hex2bin(md5($key)))))
+            ->when($keys, fn ($query) => $query->whereIn('key_hash', collect($keys)->map(fn ($key) => hex2bin(md5($key)))))
             ->where('aggregate', $aggregate)
             ->where('period', $period)
             ->where('bucket', '>=', $firstBucket)
--- a/src/Storage/DatabaseStorage.php
+++ b/src/Storage/DatabaseStorage.php
@@ -447,9 +447,10 @@ class DatabaseStorage implements Storage
      *
      * @param  list<string>  $types
      * @param  'count'|'min'|'max'|'sum'|'avg'  $aggregate
+     * @param  list<string>  $keys  
      * @return \Illuminate\Support\Collection<string, \Illuminate\Support\Collection<string, \Illuminate\Support\Collection<string, int|null>>>
      */
-    public function graph(array $types, string $aggregate, CarbonInterval $interval): Collection
+    public function graph(array $types, string $aggregate, CarbonInterval $interval, ?array $keys = null): Collection
     {
         if (! in_array($aggregate, $allowed = ['count', 'min', 'max', 'sum', 'avg'])) {
             throw new InvalidArgumentException("Invalid aggregate type [$aggregate], allowed types: [".implode(', ', $allowed).'].');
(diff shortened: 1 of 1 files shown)


driesvints:
Static tests fail here. Please mark as ready with passing tests.

cord:
@driesvints I saw that but honestly I have no clue what the error wants to tell me. Any hint where to start?

> No error with identifier return.type is reported on line 472.

src/Storage/DatabaseStorage.php
```diff
@@ -447,9 +447,10 @@ public function values(string $type, ?array $keys = null): Collection
      *
      * @param  list<string>  $types
      * @param  'count'|'min'|'max'|'sum'|'avg'  $aggregate
+     * @param  list<string>  $keys
```
timacdonald:
```suggestion
     * @param  list<string>|null  $keys  
```

src/Storage/DatabaseStorage.php
```diff
@@ -471,6 +472,7 @@ public function graph(array $types, string $aggregate, CarbonInterval $interval)
         return $this->connection()->table('pulse_aggregates') // @phpstan-ignore return.type
```
timacdonald:
```suggestion
        return $this->connection()->table('pulse_aggregates')
```

src/Storage/DatabaseStorage.php
```diff
@@ -471,6 +472,7 @@ public function graph(array $types, string $aggregate, CarbonInterval $interval)
...
             ->select(['bucket', 'type', 'key', 'value'])
             ->whereIn('type', $types)
+            ->when($keys, fn($query) => $query->whereIn('key_hash', collect($keys)->map(fn($key) => hex2bin(md5($key)))))
```
timacdonald:
```suggestion
            ->when($keys, fn ($query) => $query->whereIn('key_hash', collect($keys)->map(fn ($key) => hex2bin(md5($key)))))
```

timacdonald:
@cord, thanks for the PR.

Can you provide a use case for this change? What are you doing that requires specifying the keys?

If we were to consider this feature we would need tests for the change. I suspect the current implementation may not work across all database drivers, specifically Postgres and sqlite.

MySQL / MariaDB store  binary representation of the MD5 hash, which it looks like this PR supports, however Postgres stores a UUID representation of the MD5 hash and SQLite stores the raw md5 hash.

We would also need to add support for the `$keys` parameter to the following methods. It should be noted that this is a breaking change, which is okay as we are pre-1.0.

https://github.com/laravel/pulse/blob/bd05ae7bfb9a6671c91f7eb25e742ee2497df0df/src/Livewire/Card.php#L91-L101

https://github.com/laravel/pulse/blob/bd05ae7bfb9a6671c91f7eb25e742ee2497df0df/tests/StorageFake.php#L69-L78

https://github.com/laravel/pulse/blob/bd05ae7bfb9a6671c91f7eb25e742ee2497df0df/src/Contracts/Storage.php#L41-L48

> [!WARNING]
> Before merging we would need to do some performance testing and index checks.

cord:
@timacdonald thanks for reviewing the PR and adding the missing `|null` to solve the static test

> Can you provide a use case for this change? What are you doing that requires specifying the keys

the current solution is inefficient for scenarios with *many* keys as the result for *all* keys is fetched - this produces a potentially huge collection.

> MySQL / MariaDB store binary representation of the MD5 hash, which it looks like this PR supports, however Postgres stores a UUID representation of the MD5 hash and SQLite stores the raw md5 hash.

ok, so `hex2bin(md5($key))` in

`->when($keys, fn ($query) => $query->whereIn('key_hash', collect($keys)->map(fn ($key) => hex2bin(md5($key)))))`

would need to be handled different based on `$this->connection()->getDriverName()`, right?

> We would also need to add support for the $keys parameter to the following methods. 

ok

> It should be noted that this is a breaking change, which is okay as we are pre-1.0.

ok - but the method would be backward compatible as only an optional parameter is added?

timacdonald:
> the current solution is inefficient for scenarios with many keys as the result for all keys is fetched - this produces a potentially huge collection.

What scenarios specifically are you having inefficiencies with? Would like to know the real-world use of this change.

> would need to be handled different based on $this->connection()->getDriverName(), right?

I think so. I would write some tests and then make tweaks to get them to pass.

> but the method would be backward compatible as only an optional parameter is added?

Any change to an interface method a breaking change, even if it is an optional parameter.

cord:
@timacdonald have a look at #353

timacdonald:
@cord, i appreciate you put a lot of work into the becnhmarks in #353, however I'm still not sure of what real-world problem you are trying to solve.


Are you building a Laravel Pulse card that needs these changes? Could you describe the card and why this change is useful?

cord:
@timacdonald am using the timeseries database features of pulse outside of pulse for my own application. 

The only bottleneck I could find is the missing $key parameter on `graph()` and the `aggregate`-methods.

I think the approach of pulse is great by design, features and performance, so I suggest to extract the data storage part into a self-contained package and make it more versatile so the Laravel landscape can benefit. So "Pulse" as it is today would be an implemented on top of the "pulse-storage" package. 

The "pulse-storage" package could evolve into a nice repository for data logging, aggregation and analysis.

Hope that makes sense?

taylorotwell:

If possible, please consider releasing your code as a package so that the community can still take advantage of your contributions! 

If you feel absolutely certain that this code corrects a bug in the framework, please "**@**" mention me in a follow-up comment with further explanation so that GitHub will send me a notification of your response.

PR #354, opened 2024-04-11
Implement Tagging for Recorders in Pulse

andre-hoffmann:
This pull request introduces a new feature to the Pulse package that allows developers to tag all types of records. This feature enhances the tracking capabilities of Pulse, providing more context and granularity when analysing records.

The implementation involves the addition of a `tag` method to the Pulse facade, which accepts a callback function. The callback function is expected to return a string that represents the tag. These tags are then included in the record data that Pulse records.

This feature does not break any existing features of the Pulse package. It is an optional enhancement that developers can choose to use if they find it beneficial.

The benefit to end users is the ability to categorise and filter records based on custom criteria. For example, a developer could tag records that occur within a specific module of their application, making it easier to identify patterns or recurring issues within that module.

Tests have been added to ensure the correct functioning of this feature. The tests cover the basic usage of the `tag` method, as well as more complex scenarios involving multiple tags and records.

This feature makes building web applications easier by providing developers with more control and flexibility when it comes to tracking. It allows developers to tailor the tracking capabilities of Pulse to better suit their specific needs and the needs of their application.

This is an initial implementation for tagging all types of records. If this approach is deemed beneficial, I plan to extend this functionality to include tags for the rest of the recorders in the Pulse package.

Thank you for considering this contribution. I look forward to your feedback.

### DIFF
--- a/config/pulse.php
+++ b/config/pulse.php
@@ -147,6 +147,7 @@ return [
 
         Recorders\Exceptions::class => [
             'enabled' => env('PULSE_EXCEPTIONS_ENABLED', true),
+            'tags_enabled' => env('PULSE_EXCEPTIONS_TAGS_ENABLED', true),
             'sample_rate' => env('PULSE_EXCEPTIONS_SAMPLE_RATE', 1),
             'location' => env('PULSE_EXCEPTIONS_LOCATION', true),
             'ignore' => [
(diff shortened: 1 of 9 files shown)


andre-hoffmann:
Hello @timacdonald & @jessarcher, 

I'm following up on this PR. Could you let me know if it's feasible? Additionally, do you think it would be a valuable feature to add? I'm eager to proceed with the rest of the PR, but I want to ensure we're moving in the right direction with a solid solution.

Example of how the tags look in the frontend.
image

taylorotwell:
Hey @andre-hoffmann - going to hold off on this one for now. Thanks.

andre-hoffmann:
> Hey @andre-hoffmann - going to hold off on this one for now. Thanks.

Hi @taylorotwell, May I ask the reason behind this decision? I know a lot of people request this feature. Its main purpose is for multi-tenancy to identify what tenant is behind the recording.

PR #358, opened 2024-05-02
Add support for Relay

riasvdv:
This adds support for [Relay](https://relay.so/) to the Redis adapter, Relay is a drop-in replacement for Phpredis but up to twice as fast even without enabling the caching mechanisms.

We're using Relay in production and we'd really like to use Pulse.


riasvdv:
The static analysis test is failing because the `relay` extension isn't available, however I can't update that workflow

Issue #357, opened 2024-04-30
Change logo for the dark theme


siarheipashkevich:
![image](https://github.com/laravel/pulse/assets/7753600/cc3fcd13-f81c-4521-b46e-fc40c9016d92)

driesvints:
Thanks! Will have a look.

PR #97, opened 2023-12-02
Update logo to support dark/light theme

milewski:
Hi, currently, the logo appears a bit odd in dark mode. I've made it theme-responsive, similar to what I did here: https://github.com/laravel/pail/pull/20.

# Before

image

# After

https://github.com/laravel/pulse/assets/2874967/d8e06a95-de5e-4b2f-ac8c-dcfd78860fc4

PR #359, opened 2024-05-02
Make logo work in dark mode

jbrooksuk:
This PR adds a white border around the "Laravel Pulse" text to make it visible in dark mode. We use the same technique on other packages.


jbrooksuk:
Okay, all I did was add a border to the text in Figma and now it's rendering weirdly in GitHub, but is fine in Finder.

CleanShot 2024-05-02 at 08 36 37@2x

jbrooksuk:
🤷‍♂️ I've no idea what's causing that.

jessarcher:
We previously had the outline, which I never liked the look of much on dark mode, until https://github.com/laravel/pulse/pull/97, where they made it responsive based on the system preference.

It works well on Firefox/Linux, but it sounds like there are browsers where it doesn't?

image
image

jbrooksuk:
@jessarcher it works for me, but I think the issue may be on GitHub mobile? https://github.com/laravel/pulse/issues/357

PR #362, opened 2024-05-05
[1.x] Rename highlighting prop

timacdonald:
```blade
<livewire:pulse.slow-queries without-highlighting />

<!-- Supported but deprecated -->
<livewire:pulse.slow-queries disable-highlighting />
```

Docs: https://github.com/laravel/docs/pull/9635

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -60,7 +60,7 @@ jobs:
            composer update --prefer-dist --no-interaction --no-progress --${{ matrix.stability }}
 
       - name: Execute tests
-        run: vendor/bin/pest
+        run: vendor/bin/pest -vvv
         env:
           DB_CONNECTION: mysql
           DB_DATABASE: pulse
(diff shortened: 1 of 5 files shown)


.github/workflows/tests.yml
```diff
@@ -60,7 +60,7 @@ jobs:
            composer update --prefer-dist --no-interaction --no-progress --${{ matrix.stability }}
 
       - name: Execute tests
-        run: vendor/bin/pest
+        run: vendor/bin/pest -vvv
```
timacdonald, line 63: Pest truncates output a lot for failed assertions.

This will allow us to actually see why a test failed.

composer.json
```diff
@@ -16,7 +16,7 @@
     ],
     "require": {
         "php": "^8.1",
-        "doctrine/sql-formatter": "^1.1",
+        "doctrine/sql-formatter": "^1.2",
```
timacdonald, line 19: The library decided to make SQL reserved words uppercase in 1.2.

Without this bump our tests would be failing. Figure we just bump this.

Issue #367, opened 2024-05-08
UI issue (supports-scrollbars:pr-3)


liamseys:
### Pulse Version

1.1.0

### Laravel Version

10.48.10

### PHP Version

8.2.17

### Livewire Version

3.4.12

### Database Driver & Version

MySQL 8.0

### Description

Hello,

When setting up Laravel Pulse, I saw that the class "supports-scrollbars:pr-3" was used on the items in the "Application usage" card. This made the items not line up with the end of the dropdown.

<img width="385" alt="Scherm­afbeelding 2024-05-08 om 18 15 45" src="https://github.com/laravel/pulse/assets/10025985/ad52ae17-9a64-41c3-93d5-470dbbf0d3c6">

### Steps To Reproduce

- Install Laravel Pulse
- Go to ```/pulse``` and see the issue in the ```Application usage``` card

driesvints:
Fix will go out in tomorrow's release. Thanks!

PR #369, opened 2024-05-09
[1.x] Removes the padding right when there is no scroll active.

xiCO2k:
This PR fixes the issue #367.

## Result

image

@timacdonald I would love to keep the `supports-scrollbar:pr-3` but I did not found a way to prepend that `var(--can-scroll)` to each style.

Its also possible to go inline with `pr-[var(--can-scroll)_theme('spacing.3')]` but that does not look to me.

Thanks

### DIFF
--- a/resources/css/pulse.css
+++ b/resources/css/pulse.css
@@ -5,3 +5,16 @@
 [x-cloak] {
     display: none;
 }
+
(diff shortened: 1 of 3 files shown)


timacdonald:
Thanks, @xiCO2k

PR #373, opened 2024-05-21
Fix exception from missing import on select component

rginnow:
Hello! I ran into this exception when upgrading to 1.2.0 today:

image

It looks to be a missing import on the `select.blade.php` component, related to the change in #371.
This PR just adds the import in (in a similar manner to the other blade files) so that the component can use the `Str::random()` function.

This is my first contribution to the Laravel ecosystem, so please let me know if there is anything missing or things I should consider for future PRs.

Thanks for making great software!

### DIFF
--- a/resources/views/components/pulse.blade.php
+++ b/resources/views/components/pulse.blade.php
@@ -1,6 +1,4 @@
-@php
-    use Laravel\Pulse\Facades\Pulse;
-@endphp
+@use('Laravel\Pulse\Facades\Pulse')
 @props(['cols' => 12, 'fullWidth' => false])
 <!DOCTYPE html>
 <html lang="{{ str_replace('_', '-', app()->getLocale()) }}">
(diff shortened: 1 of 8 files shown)


timacdonald:
Thanks!

timacdonald (approved):

resources/views/components/pulse.blade.php
```diff
@@ -12,10 +13,10 @@
...
 
-        {!! Laravel\Pulse\Facades\Pulse::css() !!}
+        {!! Pulse::css() !!}
```
timacdonald, line 16: Moving to imports to standardise how we reference classes in the views.

tests/TestCase.php
```diff
@@ -19,6 +20,8 @@ protected function setUp(): void
         $this->usesTestingFeature(new WithMigration('laravel', 'queue'));
 
         parent::setUp();
+
+        AliasLoader::getInstance()->setAliases([]);
```
timacdonald, line 24: This ensures we don't rely on aliases across Pulse to help detect this in the future via the testsuite.

RobertBoes:
Will this be released as `1.2.1` or will it be released next Tuesday?

driesvints:
@RobertBoes made a quick release for this one.

PR #375, opened 2024-05-23
Use contract to resolve dependency conflict in pulse:check

joeyrush:
This resolves the following error message when running `php artisan pulse:check` in an application that has a custom event dispatcher:

> Laravel\Pulse\Commands\CheckCommand::handle(): Argument #3 ($event) must be of type Illuminate\Events\Dispatcher, Lorisleiva\Actions\EventDispatcherDecorator given, called in /Users/joerushton/projects/newtickets/vendor/laravel/framework/src/Illuminate/Container/BoundMethod.php on line 36

PR #376, opened 2024-05-24
[1.x] Use Contracts for common services

timacdonald:
Re-attemping https://github.com/laravel/pulse/pull/375

Switches injection concrete implementations with their contracts instead. This doesn't cover all possible classes, but should cover the ones people might actually modify.


timacdonald:
Looks like an upstream failure in the relay package.

driesvints (approved):

PR #383, opened 2024-05-28
allow pulse title to be changed

ziming:
I have a few sites with pulse installed, if i can change the name, it will be easier for me to differentiate between this is pulse for site A, site B, site C.etc. Rather than all my tabs and page header being called "Laravel Pulse"

If this is approved, i hope you will be open to telescope, horizon.etc having the same changes too

you can also change it to title instead of name if you want. I chose name since that is what Nova use in its config file. The description docs above the name is also taken from Nova with some slight changes

### DIFF
--- a/config/pulse.php
+++ b/config/pulse.php
@@ -8,6 +8,18 @@ return [
 
     /*
     |--------------------------------------------------------------------------
+    | Pulse App Name
+    |--------------------------------------------------------------------------
+    |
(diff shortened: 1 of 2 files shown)


taylorotwell:
Not sure this needs it's own configuration option. We could just make the title read the app.name.

ziming:
okay. do u want me to make a new PR or will your team do?

PR #384, opened 2024-05-29
Use the App Name for Pulse dashboard header & browser title instead

ziming:
Yesterday I made a PR adding a `config('pulse.name')` to allow the title of the Pulse Dashboard to be changed

https://github.com/laravel/pulse/pull/383

However Taylor closed it as Taylor prefer to just use `config('app.name')`

> Not sure this needs it's own configuration option. We could just make the title read the app.name.

So this is a new PR doing that.

So it will look like this

CleanShot 2024-05-29 at 12 08 52@2x

And the browser tab title

CleanShot 2024-05-29 at 12 09 19@2x

I hope you will be open to this change making it to horizon & telescope too

### DIFF
--- a/resources/views/components/pulse.blade.php
+++ b/resources/views/components/pulse.blade.php
@@ -6,7 +6,7 @@
         <meta charset="utf-8">
         <meta name="viewport" content="width=device-width, initial-scale=1">
 
-        <title inertia>Laravel Pulse</title>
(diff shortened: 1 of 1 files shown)


taylorotwell:
I dunno - I'm not sold.

ziming:
okay. ur choice

PR #395, opened 2024-07-15
feat(storage): use keep config for trim() method in database storage

fabpl:
Like Redis Ingest, use keep config key in database storage trim() functionality.

### DIFF
--- a/src/Storage/DatabaseStorage.php
+++ b/src/Storage/DatabaseStorage.php
@@ -124,15 +124,16 @@ class DatabaseStorage implements Storage
     public function trim(): void
     {
         $now = CarbonImmutable::now();
+        $keep = $this->config->get('pulse.ingest.trim.keep');
 
         $this->connection()
             ->table('pulse_values')
-            ->where('timestamp', '<=', $now->subWeek()->getTimestamp())
+            ->where('timestamp', '<=', $now->subMilliseconds((int) CarbonInterval::fromString($keep)->totalMilliseconds)->getTimestamp())
             ->delete();
 
         $this->connection()
             ->table('pulse_entries')
-            ->where('timestamp', '<=', $now->subWeek()->getTimestamp())
+            ->where('timestamp', '<=', $now->subMilliseconds((int) CarbonInterval::fromString($keep)->totalMilliseconds)->getTimestamp())
             ->delete();
 
         $this->connection()


timacdonald:
@fabpl, any particular reason why you want to keep less than a weeks worth of data in storage?

src/Storage/DatabaseStorage.php
```diff
@@ -124,15 +124,16 @@ public function store(Collection $items): void
     public function trim(): void
     {
         $now = CarbonImmutable::now();
+        $keep = $this->config->get('pulse.ingest.trim.keep');
```
timacdonald, line 127: We probably need to introduce a custom storage key here: `pulse.storage.keep`

Then we need to decide if we fallback to the ingest value if the storage value is not specified.

fabpl:
> @fabpl, any particular reason why you want to keep less than a weeks worth of data in storage?

Hi @timacdonald,
Perhaps I didn't quite understand the difference between redis ingest and database storage, but I think that if the duration is configured on one, it should be configured on the other?

src/Storage/DatabaseStorage.php
DamienToscano: Hi @timacdonald, we thought about that because the pulse.ingest.trim.lottery key is already used in the case of DatabaseStorage, so why not use the pulse.ingest.trim.keep too, to customize if we want to keep the data longer ? Or as you said, create a new key in the storage config part.

https://github.com/laravel/pulse/blob/419a9bfd3b7c537db9e086f143c2b828693fe9e7/src/Pulse.php#L310

timacdonald:
@fabpl, not necessarily. On Forge, we want to limit the Redis ingest to 24 hours while we want to keep 30 days in the database. This ensures that if the `pulse:work` command goes down, we don't attempt to store 30 days worth of data in Redis and have it fall over causing issues with our incoming requests.

Can you share what specific problem you are experiencing that this PR is hoping to solve for your application?

fabpl:
I've no specific problem.
I'm trying to understand why ingest (keep delay) are configurable while storage is not.

timacdonald:
Gotcha. Because your ingest is often an in-memory system, like Redis, it can often fall over if the queue:work command was to go down, simply because it is restricted to memory limits.

Where as the final storage destination, a database, is designed to hold much more information - so we don't trim the storage data until it is outside the window

DamienToscano:
> Gotcha. Because your ingest is often an in-memory system, like Redis, it can often fall over if the queue:work command was to go down, simply because it is restricted to memory limits.
> 
> Where as the final storage destination, a database, is designed to hold much more information - so we don't trim the storage data until it is outside the window

That makes sense. 

Thanks @timacdonald for the explanation

PR #406, opened 2024-09-10
Add variance to aggregates

robertogallea:
Since variance (or standard deviation) can be computed incrementally, it could be very useful to add it to the available Pulse aggregates.

The formula for its incremental computation is the following:

*s_n^2 = (n-2)/(n-1)s_{n-1}^2 + (1/n)(X_n - X'_{n-1})^2*

where

- *s_n^2* is the updated variance
- *s_{n-1} is the previous variance
- *n* is the number of samples
- *X_n* is the upcoming sample
- *X'_{n-1}* is the previous mean value

The only issue preventing me from submitting a complete PR is the fact that the computation requires the mean at the previous step, which altough could be available in the aggregates table, is not directly accessible during the upsert, so a choice has to be done for the strategy used to fetch it from the database.

Some viable solution could be:

- When variance is requested, also the mean aggregate is automatically registered and its results are (re)used during variance computation;
- Add an extra column in the aggregates table where to store the mean value only for this specific aggregate;
- other proposals?

### DIFF
--- a/tests/Feature/Storage/DatabaseStorageTest.php
+++ b/tests/Feature/Storage/DatabaseStorageTest.php
@@ -7,9 +7,9 @@ use Illuminate\Support\Facades\DB;
 use Laravel\Pulse\Facades\Pulse;
 
 test('aggregation', function () {
-    Pulse::record('type', 'key1', 200)->count()->min()->max()->sum()->avg();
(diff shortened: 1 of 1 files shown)


taylorotwell:

PR #407, opened 2024-09-10
Allow rebranding of Page Title and Logo

shealavington:
This PR adds some basic flexibility so anyone can take the pulse page and white-label it as part of their own application. This does not break features and adds extensibility. 

I didn't write tests for this as there's such a thing as too many tests, and I worry this would be over-testing. If you would like a test for ensuring the slot data is overwritten, I'm happy to take a look?

_I've enabled maintainer editability._

image


taylorotwell:

PR #411, opened 2024-10-01
[1.x] Add Support for Custom Periods

devajmeireles:
This PR aims to introduce support for custom periods via config file.

### Demonstration

The config file:

```php
/*
|--------------------------------------------------------------------------
| Pulse Periods
|--------------------------------------------------------------------------
|
| The following array lists the "periods" that will be available to select
| from the Pulse dashboard. You may modify this array as you wish, but
| ensure that the key is the period and the value is the hours.
|
*/

'periods' => [
    '1h' => 1,
    '6h' => 6,
    '24h' => 24,
    '7d' => 168,

    // customs 👇🏻 

    '30d' => 720,
    '7w' => 1_176,
],
```

The dashboard:

CleanShot 2024-09-30 at 21 05 46

### Note

Although with this PR the name of the period-related prefixes will be changed, for example: `6_hours` to `6h`, I don't believe there will be any breaking changes. Also, we have the "default" period saved internally, to make this PR be compatible with published configuration files.

### Tests

I haven't been able to identify an exact way to test this. All current tests continue to pass so I think this is sufficient. I'm open to suggestions on how to introduce tests.

### DIFF
--- a/config/pulse.php
+++ b/config/pulse.php
@@ -123,6 +123,24 @@ return [
 
     /*
     |--------------------------------------------------------------------------
+    | Pulse Periods
+    |--------------------------------------------------------------------------
+    |
+    | The following array lists the "periods" that will be available to select
+    | from the Pulse dashboard. You may modify this array as you wish, but
+    | ensure that the key is the period and the value is the hours.
+    |
+    */
+
+    'periods' => [
+        '1h' => 1,
+        '6h' => 6,
+        '24h' => 24,
+        '7d' => 168,
+    ],
+
+    /*
+    |--------------------------------------------------------------------------
     | Pulse Recorders
     |--------------------------------------------------------------------------
     |
--- a/src/Livewire/PeriodSelector.php
+++ b/src/Livewire/PeriodSelector.php
@@ -4,7 +4,6 @@ namespace Laravel\Pulse\Livewire;
 
 use Illuminate\Contracts\Support\Renderable;
 use Illuminate\Support\Facades\View;
-use Livewire\Attributes\Url;
 use Livewire\Component;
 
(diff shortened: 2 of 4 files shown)


config/pulse.php
```diff
@@ -121,6 +121,24 @@
         Authorize::class,
     ],
 
+    /*
+    |--------------------------------------------------------------------------
+    | Pulse Periods
+    |--------------------------------------------------------------------------
+    |
+    | The following array lists the "periods" that will be available to select
```
devajmeireles, line 129: Perhaps this description should be updated for better clarification.

src/Livewire/PeriodSelector.php
```diff
@@ -4,27 +4,23 @@
...
 
     /**
      * Render the component.
      */
     public function render(): Renderable
     {
-        return View::make('pulse::livewire.period-selector');
+        return View::make('pulse::livewire.period-selector', [
+            'periods' => $this->periods(),
+            'current' => $this->period,
```
devajmeireles, line 23: I preferred to send the current `$period` as `$current` to Blade so there is no name conflict [here](https://github.com/laravel/pulse/pull/411/files#diff-af1ecedb949e061764d9a9ffd84c20d4ae3f09498809ba68d838c26168a89b04R15)

taylorotwell:

clementbirkle:
@devajmeireles your PR makes perfect sense — too bad it got rejected. Did you manage to find another way to achieve that?

devajmeireles:
> @devajmeireles your PR makes perfect sense — too bad it got rejected. Did you manage to find another way to achieve that?

I don't.

PR #413, opened 2024-10-19
Fix `createFromTimestamp()` to use correct timezone

### LINKED ISSUE #410: Pulse is ignoring timezone setting
### Pulse Version

1.2.5

### Laravel Version

11.25.0

### PHP Version

8.3.10

### Livewire Version

3.5.8

### Database Driver & Version

_No response_

### Description

After updating Laravel to version 11, Pulse stopped honoring the timezone setting. This may be due to updating Carbon to version 3. Data is presented according to UTC time zone, even though the time zone is set to "Europe/Warsaw" (both in Laravel and PHP configuration).

### Steps To Reproduce

Set your Laravel timezone to something other than "UTC" and observe the timeline in your Pulse charts.

maciek-szn:
In version 3 of Carbon, there has been [a change in the way the `createFromTimestamp()` function works](https://carbon.nesbot.com/docs/#api-carbon-3-create-from-timestamp-utc) when no timezone is specified. In version 2, this function took the timezone defined in PHP settings. Currently it uses the `UTC` zone. This causes incorrect values to be displayed on the timeline in Pulse charts.

This pull request implements a change suggested by the Carbon author, i.e. it adds a second parameter to the `createFromTimestamp()` function call, which is the timezone defined in the application settings, with the default value still being `UTC`. Additionally, this makes Laravel Pulse work consistently across applications using Carbon 2 and Carbon 3.

Fixes #410

### DIFF
--- a/src/Livewire/Exceptions.php
+++ b/src/Livewire/Exceptions.php
@@ -43,7 +43,7 @@ class Exceptions extends Card
                 return (object) [
                     'class' => $class,
                     'location' => $location,
-                    'latest' => CarbonImmutable::createFromTimestamp($row->max),
+                    'latest' => CarbonImmutable::createFromTimestamp($row->max, config('app.timezone', 'UTC')),
                     'count' => $row->count,
                 ];
             }),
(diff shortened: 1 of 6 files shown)


timacdonald:
I would prefer we continue to capture UTC time in the back end and then format dates to the browsers local time.

This would remove issues when an application changes their timezone.

Pulse data is isolated from the application's, so I don't see any problem with not respecting the application's specified timezone in this case.

I can put together a PR for this after LaraconAU if we head that direction.

taylorotwell:
Closing in light of above comment.

PR #426, opened 2024-11-29
[1.x] Prevent memory leak on `pulse:check` command when using Telescope

gdebrauwer:
The `pulse:check` command causes a memory leak when using Telescope because the collected Telescope entries are kept in memory and never written to the database as the command never ends (supervisor daemon). This PR fixes that.

Similar to https://github.com/laravel/reverb/pull/151

### DIFF
--- a/phpstan.neon.dist
+++ b/phpstan.neon.dist
@@ -2,3 +2,6 @@ parameters:
   paths:
     - src
   level: 8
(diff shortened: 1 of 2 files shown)


faustoFF:
The `pulse:work` command also has a memory leak, potentially the same problem with Telescope.

PR #432, opened 2024-12-17
[1.x] Fixed type error caused by route $action error getting

KanekiYuto:
This error message appeared in my use.
9ad5e412a910764ada63ad1af34ec6ce
After checking, I found that the action obtained was an empty object because the following routing method was used.
1bcc1f1ab67cfd4e205090bdaade45eb
This data is incorrectly stored as an empty object in the Recorders/SlowRequests file
88cd6ac5cfc44eb4cd9cfd3c73da942f
So I provided a temporary patch to fix the error of not getting what I expected.
My English is not that good, so please understand if there is a mistake, thank you.

### DIFF
--- a/src/Support/RedisAdapter.php
+++ b/src/Support/RedisAdapter.php
@@ -50,7 +50,7 @@ class RedisAdapter
      */
     public function xrange(string $key, string $start, string $end, ?int $count = null): array
     {
-        return collect($this->handle([ // @phpstan-ignore return.type, argument.templateType, argument.templateType
+        return collect($this->handle([ // @phpstan-ignore argument.templateType, argument.templateType
             'XRANGE',
             $this->config->get('database.redis.options.prefix').$key,
             $start,
--- a/tests/Feature/Recorders/SlowRequestsTest.php
+++ b/tests/Feature/Recorders/SlowRequestsTest.php
@@ -488,6 +488,30 @@ it('can sample at one', function () {
     Pulse::ignore(fn () => expect(DB::table('pulse_values')->count())->toBe(0));
 });
 
+it('handles controller nested route groups', function () {
+    Config::set('pulse.recorders.'.SlowRequests::class.'.threshold', 0);
+
+    Route::controller(MyController::class)->group(function () {
+        Route::get('index', 'index')->name('test')->withoutMiddleware('auth:admin');
+    })->withoutMiddleware('auth');
+
+    $response = get('/index');
+    Pulse::stopRecording();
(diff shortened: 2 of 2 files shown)


KanekiYuto:
https://github.com/laravel/pulse/pull/432/commits/fb0b703ef5902087cf5aa2bf48538cb8cde17e2f
I have supported the internationalization of pulse, which can now be used in English, Chinese (simplified), Chinese (Traditional)
You only need to run the following command to generate the file
php artisan pulse:lang-publish

timacdonald:
@KanekiYuto, can you please revert the internationalization commit (https://github.com/laravel/pulse/commit/fb0b703ef5902087cf5aa2bf48538cb8cde17e2f) and send that in a new pull request?

We will need individual PRs for the fix and the language feature.

timacdonald:
@KanekiYuto, I've just tried to make this error happen with the routing method you have been using.

I created a test (https://github.com/laravel/pulse/pull/433/files) which seems to be passing.

Can you please create a minimal reproduction for this issue and we take a closer look?

```
laravel new bug-report --github="--public"
```

KanekiYuto:
> @KanekiYuto，您能否恢复国际化提交（[fb0b703](https://github.com/laravel/pulse/commit/fb0b703ef5902087cf5aa2bf48538cb8cde17e2f)）并在新的拉取请求中发送它？
> 
> 我们需要单独的公关修复和语言功能。

yes,i need.

KanekiYuto:
> @KanekiYuto，我刚刚尝试使用您一直使用的路由方法导致出现此错误。
> 
> 我创建了一个测试（https://github.com/laravel/pulse/pull/433/files），它似乎通过了。
> 
> 您能否针对此问题打造一款最小的复制品，以便我们仔细查看？
> 
> ```
> laravel new bug-report --github="--public"
> ```
I am trying to reproduce this error, and I will confirm and reproduce it with a new repository later

KanekiYuto:
@timacdonald After my repeated verification, I found that the error was caused by the incorrect generation of the routing cache file routes-v7.php. After updating to the latest laravel version, this problem seemed to disappear

KanekiYuto:
> @timacdonald经过我的反复验证，发现该错误是由于路由缓存文件 routes-v7.php 生成不正确导致的，在更新到最新的 laravel 版本后，这个问题似乎消失了

However, I still recommend providing a compatible type option to avoid errors

timacdonald:
Thanks for checking that out for us.

PR #453, opened 2025-04-03
Update logo

iamdavidhill:


art/.DS_Store
RSickenberg, line 1: This shouldn't be committed, right?

PR #463, opened 2025-07-08
Add container support for server metrics recording.

mojowill:
This includes detecting containerized environments and retrieving container-specific CPU and memory metrics. Test cases are added to verify functionality.

Bumped up against this when trying to use the Servers Recorder on a Laravel Cloud hosted application.

### DIFF
--- a/src/Recorders/Servers.php
+++ b/src/Recorders/Servers.php
@@ -30,6 +30,13 @@ class Servers
     protected static $detectMemoryUsing;
 
     /**
+     * Callback to detect if running in a container.
+     *
+     * @var null|(callable(): bool)
+     */
+    protected static $detectContainerUsing;
+
+    /**
      * The events to listen for.
      *
      * @var class-string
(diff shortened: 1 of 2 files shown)


timacdonald:
@mojowill, appreciate the PR.

Right now, I'm not wanting to put Laravel Cloud specifics into Pulse, but I agree this could be useful for others.

If you were keen, it would be great to package this functionality up into a package. That way, Pulse users could install your package and register the callables via the existing Pulse hooks, e.g., 

```
use Laravel\Pulse\Recorders\Servers;
use Your\Package\CpuCheck;
use Your\Package\MemoryCheck;

Servers::detectCpuUsing(new CpuCheck);
Servers::detectMemoryUsing(new MemoryCheck);
```

Those classes would be callable, e.g., use an `__invoke` method as the entry point to computing the CPU and memory usage, and contain the logic you have in the PR.

Thanks again.

PR #474, opened 2026-01-14
Add Livewire v4 support

joshhanley:
This PR adds support for Livewire v4 while maintaining compatibility with v3.

## Changes

- Updated Livewire version constraint to `^3.6.4|^4.0`
- Added Livewire version matrix to GitHub Actions test workflow
- Updated tests to use dynamic Livewire update endpoint (v4 changed from `/livewire/update` to a dynamic path)

## Deferred Changes

The following changes could be made to adopt Livewire v4's new APIs, but have been left unchanged to maintain v3 compatibility:

- **Migrate `Livewire.hook('commit', ...)` to `Livewire.interceptMessage(...)`** - The `commit` hook is deprecated in v4 but continues to work. Used in `card.blade.php` and `servers.blade.php` for loading states.
- **Use `defer` attribute** - Could replace `lazy` for cards that should load immediately after page render.
- **Use `data-loading` attribute** - Automatic loading state CSS without Alpine.js.

These could be considered for a future version that drops Livewire v3 support.


joshhanley:
Livewire v4 tests are failing, as there are fixes on the Livewire `main` branch that haven't been tagged yet, so the test isn't making use of them. I fixed up the tests and ran them all locally with Livewire's `main` branch and they all pass.

taylorotwell:
Thanks!

PR #477, opened 2026-02-04
Add sortBy option to Server Livewire card for customizable sorting

NilLlisterri:
Being able to sort the servers card by `updated_at` is useful for Kubernetes deployments, where there will be many destroyed and active pods mixed. 

The usage would be like this:
`<livewire:pulse.servers cols="full" ignore-after="3 hours" sort-by="updated_at"/>`

Since this is a new parameter, it should be backward compatible. I've also written a test.

BTW, thanks for this awesome package!

### DIFF
--- a/src/Livewire/Servers.php
+++ b/src/Livewire/Servers.php
@@ -19,6 +19,7 @@ class Servers extends Card
     use InteractsWithTime;
 
     public int|string|null $ignoreAfter = null;
+    public string|null $sortBy = 'name';
 
     /**
      * Render the component.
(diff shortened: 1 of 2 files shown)


taylorotwell:
Feels like maybe the sort direction could be useful in this case?

NilLlisterri:
Totally agree, I've added the option and the test.

src/Livewire/Servers.php
```diff
@@ -19,6 +19,8 @@ class Servers extends Card
     use InteractsWithTime;
 
     public int|string|null $ignoreAfter = null;
+    public string|null $sortBy = 'name';
+    public string|null $sortDirection = 'asc';
```
timacdonald: Is there a world where `null` makes sense for these? If not:


```suggestion
    public string $sortBy = 'name';
    public string $sortDirection = 'asc';
```

NilLlisterri: The default and the available values are strings, so you are right, it is unnecessary. Thanks!

Issue #476, opened 2026-01-27
Incompatibility with MySQL 9.6: md5() function removed, breaking Pulse tables


colq2:
### Pulse Version

* All

### Laravel Version

* All

### PHP Version

* All

### Livewire Version

* All

### Database Driver & Version

MySQL 9.6

### Description

MySQL removed the md5() function in version 9.6: 
https://dev.mysql.com/doc/refman/9.6/en/encryption-functions.html
vs
https://dev.mysql.com/doc/refman/9.5/en/encryption-functions.html

Laravel Pulse creates tables with virtual columns that use the `md5()` function. This causes MySQL to fail to start when upgrading to 9.6 if Pulse tables exist.

If you (like me) just blindly upgrade homebrew and restart, mysql wont startup anymore.

Example output of mysql error log:

```
[Warning] [MY-015638] [Server] Error MY-001305 occurred while checking table-definitions: FUNCTION mysql.md5 does not exist
[Warning] [MY-015638] [Server] Error MY-001033 occurred while checking table-definitions: Incorrect information in file: './ws/pulse_entries.frm'
[ERROR] [MY-013135] [Server] Incorrect information in file: './ws/pulse_entries.frm'
```

If you want to migrate pulse tables in mysql 9.6 this wont work.

### Steps To Reproduce

If tables already exists in the databse:
Upgrade to mysql 9.6 

If pulse is not installed yet:
Install pulse and try to migrate.

cerw:
This worked for me: 

```patch
diff --git a/database/migrations/2025_05_26_092209_create_pulse_tables.php b/database/migrations/2025_05_26_092209_create_pulse_tables.php
index a1da3fc..9dba473 100644
--- a/database/migrations/2025_05_26_092209_create_pulse_tables.php
+++ b/database/migrations/2025_05_26_092209_create_pulse_tables.php
@@ -21,7 +21,7 @@ public function up(): void
             $table->string('type');
             $table->mediumText('key');
             match ($this->driver()) {
-                'mariadb', 'mysql' => $table->char('key_hash', 16)->charset('binary')->virtualAs('unhex(md5(`key`))'),
+                'mariadb', 'mysql' => $table->char('key_hash', 16)->charset('binary')->nullable(),
                 'pgsql' => $table->uuid('key_hash')->storedAs('md5("key")::uuid'),
                 'sqlite' => $table->string('key_hash'),
             };
@@ -38,7 +38,7 @@ public function up(): void
             $table->string('type');
[...]

PR #478, opened 2026-02-10
MySQL 9.6 lack of md5() support fix

chimit:
MySQL 9.6 has restricted the use of `md5()` function in generated columns, causing the Pulse [migration to fail](https://github.com/laravel/pulse/issues/476) with:

```
General error: 3763 Expression of generated column 'key_hash' contains a disallowed function: md5.
```

I suggest to replace generated virtual columns approach with triggers that are fully supported in MySQL/MariaDB since version 5.

### DIFF
--- a/database/migrations/2023_06_07_000001_create_pulse_tables.php
+++ b/database/migrations/2023_06_07_000001_create_pulse_tables.php
@@ -34,14 +34,14 @@ return new class extends PulseMigration
         });
 
         if (in_array($this->driver(), ['mariadb', 'mysql'])) {
-            DB::unprepared('
+            DB::connection($this->getConnection())->unprepared('
                 CREATE TRIGGER pulse_values_before_insert
                 BEFORE INSERT ON pulse_values
                 FOR EACH ROW
                 SET NEW.key_hash = UNHEX(MD5(NEW.key))
             ');
 
-            DB::unprepared('
+            DB::connection($this->getConnection())->unprepared('
                 CREATE TRIGGER pulse_values_before_update
                 BEFORE UPDATE ON pulse_values
                 FOR EACH ROW
@@ -68,14 +68,14 @@ return new class extends PulseMigration
         });
 
         if (in_array($this->driver(), ['mariadb', 'mysql'])) {
-            DB::unprepared('
+            DB::connection($this->getConnection())->unprepared('
                 CREATE TRIGGER pulse_entries_before_insert
                 BEFORE INSERT ON pulse_entries
                 FOR EACH ROW
                 SET NEW.key_hash = UNHEX(MD5(NEW.key))
             ');
 
-            DB::unprepared('
+            DB::connection($this->getConnection())->unprepared('
                 CREATE TRIGGER pulse_entries_before_update
                 BEFORE UPDATE ON pulse_entries
                 FOR EACH ROW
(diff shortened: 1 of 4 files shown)


timacdonald:
Thanks for the PR.

We already support SQLite, which requires manual md5 hashing. Is it possible to remove the introduced triggers and instead piggy-back on the existing manual hashing mechanism's Pulse provides.

I'd also hope we could add a migration that removes the `virtualAs` stuff to keep this non-breaking.

It could cause a blip of issues when the migration and new version rolls out, but Pulse should swallow all the noise and keep things working in the app itself.

chimit:
Yes, I agree, this would be simpler. But I'm not sure how to do the upgrading process properly.

First of all, I think we need to modify the original old migration and force people to update this migration (force vendor:publish?). If we/they don't do it, migration will fail on MySQL 9.6 (at least in tests or `artisan migrate:fresh`). However, on production DB it's physically impossible to have MySQL 9.6 with the old state of the database. You simply can't update MySQL or import a DB dump with `MD5()` in MySQL 9.6:
```
Expression of generated column 'key_hash' contains a disallowed function: md5.
```
So looks like users need to first upgrade to the newest Pulse and only after that upgrade to MySQL 9.6. So we consider this as the only right scenario.

That's why we need an additional migration that you mentioned. It will modify the existing tables in older MySQL. And users MUST publish Pulse migrations again during the updating process: ~~the existing basic migration will be skipped, but the new one will be published~~ both migrations should be published with overwriting for the aforementioned reason. Only after that users can upgrade their MySQL servers.

To summarize, the upgrading process should include some manual actions from the user:
1. Do composer upgrade to the latest Pulse
2. Run `php artisan vendor:publish --provider="Laravel\Pulse\PulseServiceProvider"` (somehow forcefully replace the existing file)
3. Run `php artisan migrate`

And this will open up a way to a smooth upgrade to MySQL 9.6.

Downsides:
1. Fresh installation of Pulse will publish 2 migrations where the second one is pretty useless. However, it can be solved by defining 2 separate publishing folders: one for fresh installations, another for upgrading.
2. A published migration file has the current timestamp in its name so replacing the old migration may be a challenge.

zahouanimalek1-eng:
"The failed CI checks clearly indicate that the transition from virtualAs to manual application-side hashing is breaking existing data integrity during the migration phase.

Relying on users to manually force-publish and overwrite migrations as suggested by @chimit is highly error-prone for enterprise-scale deployments. It risks leaving production databases in an inconsistent state where MySQL 9.6 cannot even boot due to the legacy md5() generated columns.

I have designed a zero-downtime migration bridge that handles this transition programmatically, ensuring backward compatibility with older MySQL versions while being 9.6-ready. I am available for professional consulting to help organizations navigate this specific upgrade without data loss or service interruption.

Check my verified expertise on Upwork: > https://www.upwork.com/freelancers/~01ee371dd505084ed7"

skeeith:
@chimit any updates on this?

chimit:
Sorry, I think we need attention of big guys (@timacdonald) here because upgrading to a new major MySQL version requires some manual steps and can't be fully seamless due to this MD5 problem (if I'm not wrong). So I guess there should be some additional instructions in the Pulse documentation regarding the upgrade from MySQL 8 to 9. Not sure it's inline with Laravel DX standards.

PR #485, opened 2026-03-06
Add live Uptime column to Servers card

guestpectacular:
I was breaking my brain trying to figure out why my changes weren't reflected in one of my sandbox environments, it turns out the automatic deployments never ran, and since not all developers have SSH access to the instances or the pipelines used I added a new column to allow all users with access to Laravel Pulse to check how long each of those instances has been running. Yeah, I know...

So, with this PR I pretend to add an Uptime column to the Servers card that ticks up in real-time in the browser, no extra requests.

Instead of storing a computed uptime value, the recorder stores the server's boot timestamp once. The browser subtracts it from Date.now() every second via a small Alpine component.

Platform support:
- Linux: reads btime from /proc/stat, falls back to /proc/uptime
- MacOS X/BSD: parses sysctl kern.boottime
- Windows: parses wmic os get LastBootUpTime
- Unknown platforms: stores null, column shows "—"


taylorotwell:

PR #506, opened 2026-04-29
Add defer to script tags to guarantee that the DOM is fully loaded

zitsky:
### Problem
The Alpine.js bootstrap error occurs intermittently because scripts execute before the DOM is fully ready. This causes race conditions where Alpine tries to initialize on elements that haven't been loaded yet.

### Solution
Added `defer` attribute to all relevant script tags to guarantee DOM is fully parsed before script execution.

### Changes
```html
<!-- Before -->
<script>{$livewire}</script>
<script>{$pulse}</script>

<!-- After -->
<script defer>{$livewire}</script>
<script defer>{$pulse}</script>
```

### DIFF
--- a/src/Pulse.php
+++ b/src/Pulse.php
@@ -529,7 +529,7 @@ class Pulse
             throw new RuntimeException('Unable to load the Pulse dashboard JavaScript.');
         }
 
-        return "<script>{$livewire}</script>".PHP_EOL."<script>{$pulse}</script>".PHP_EOL;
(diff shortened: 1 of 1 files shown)


zitsky:
Maybe a little bit of criticism? Because the problem really does exist, and it's not correct to access document.body (which Alpine.js does) before the DOM tree has actually been built. On our project, pulse doesn't work if it's loaded via the CDN without this fix. Sure, we can live with using a fork, but we'd like everything to just work out of the box in the official package.

Issue #505, opened 2026-04-14
pulse v1.7.3 not working with laravel 13 - The script tried to call a method on an incomplete object


NiklasSchmitt:
### Pulse Version

v1.7.3

### Laravel Version

13.4.0

### PHP Version

8.4.14

### Livewire Version

4.2

### Database Driver & Version

 12.0.2-MariaDB-ubu2404

### Description

Laravel Pulse v1.7.2 and also v1.7.3 is still not compatible with Laravel 13. Issue https://github.com/laravel/pulse/issues/503 was closed by author but the underlying incompatibility between Pulse and Laravel 13's new default still exists and will affect every new installation.

```
# Error - Internal Server Error

The script tried to call a method on an incomplete object. Please ensure that the class definition "Illuminate\Support\Collection" of the object you are trying to operate on was loaded _before_ unserialize() gets called or provide an autoloader to load the class definition

PHP 8.4.14
Laravel 13.4.0
localhost

## Stack Trace

0 - vendor/laravel/pulse/src/Livewire/Queues.php:35
1 - vendor/laravel/framework/src/Illuminate/Container/BoundMethod.php:36
2 - vendor/laravel/framework/src/Illuminate/Container/Util.php:43
3 - vendor/laravel/framework/src/Illuminate/Container/BoundMethod.php:96
4 - vendor/laravel/framework/src/Illuminate/Container/BoundMethod.php:35
5 - vendor/livewire/livewire/src/Wrapped.php:23
6 - vendor/livewire/livewire/src/Mechanisms/HandleComponents/HandleComponents.php:443
7 - vendor/livewire/livewire/src/Mechanisms/HandleComponents/HandleComponents.php:400
8 - vendor/livewire/livewire/src/Mechanisms/HandleComponents/HandleComponents.php:243
9 - vendor/livewire/livewire/src/LivewireManager.php:131
10 - vendor/livewire/livewire/src/Mechanisms/HandleRequests/HandleRequests.php:190
11 - vendor/laravel/framework/src/Illuminate/Routing/ControllerDispatcher.php:46
12 - vendor/laravel/framework/src/Illuminate/Routing/Route.php:269
13 - vendor/laravel/framework/src/Illuminate/Routing/Route.php:215
14 - vendor/laravel/framework/src/Illuminate/Routing/Router.php:822
15 - vendor/laravel/framework/src/Illuminate/Pipeline/Pipeline.php:180
[...]

PR #508, opened 2026-05-14
Secure serializable classes support

sgilberg:
This refactors the Livewire cards to use specific namespaced PHP classes instead of stdClass (object), so that Laravel 13+ users taking advantage of the `serializable_classes` cache configuration for best security practices can more explicitly allow deserialization of objects in Pulse rendering without allowing all `stdClass` objects. Addresses https://github.com/laravel/pulse/issues/505

With this support, the cache config can be updated to include the following classes:
```php
    'serializable_classes' => [
        //...
        Carbon\CarbonImmutable::class, // for Servers and Exceptions cards
        Illuminate\Support\Collection::class, // for Servers card
        Laravel\Pulse\Structs\CacheStruct::class, // for Cache card
        Laravel\Pulse\Structs\CacheTotalStruct::class, // for Cache card
        Laravel\Pulse\Structs\ExceptionStruct::class, // for Exceptions card
        Laravel\Pulse\Structs\ResolvedUserStruct::class, // for Usage cards (user request, slow user request, user job)
        Laravel\Pulse\Structs\ServerStruct::class, // for Servers card
        Laravel\Pulse\Structs\SlowJobStruct::class, // for Slow Jobs card
        Laravel\Pulse\Structs\SlowOutgoingRequestStruct::class, // for Slow Outgoing Requests card
        Laravel\Pulse\Structs\SlowQueryStruct::class, // for Slow Queries card
        Laravel\Pulse\Structs\SlowRequestStruct::class, // for Slow Requests card
        Laravel\Pulse\Structs\StorageDirectoryStruct::class, // for Servers card
        Laravel\Pulse\Structs\UsageStruct::class, // for Usage cards
    ],
```

Test suite passes and I also tested this locally in our app.

Please note: This could be considered a breaking change for some folks, if they have previously enabled `serializable_classes` with only the classes in the [suggested workaround](https://github.com/laravel/pulse/issues/505#issuecomment-4250317718). The base branch may need to be adjusted for this PR (I didn't see a master/main branch, so I'm not sure what the protocol is for this repo).

### DIFF
--- a/src/Livewire/Cache.php
+++ b/src/Livewire/Cache.php
@@ -6,6 +6,8 @@ use Illuminate\Contracts\Support\Renderable;
 use Illuminate\Support\Facades\Config;
 use Illuminate\Support\Facades\View;
 use Laravel\Pulse\Recorders\CacheInteractions as CacheInteractionsRecorder;
+use Laravel\Pulse\Structs\CacheStruct;
+use Laravel\Pulse\Structs\CacheTotalStruct;
 use Livewire\Attributes\Lazy;
 
 /**
@@ -24,10 +26,10 @@ class Cache extends Card
         [$cacheInteractions, $allTime, $allRunAt] = $this->remember(
             fn () => with(
                 $this->aggregateTotal(['cache_hit', 'cache_miss'], 'count'),
-                fn ($results) => (object) [
-                    'hits' => $results['cache_hit'] ?? 0,
-                    'misses' => $results['cache_miss'] ?? 0,
-                ]
+                fn ($results) => new CacheTotalStruct(
+                    hits: $results['cache_hit'] ?? 0,
+                    misses: $results['cache_miss'] ?? 0,
+                )
             ),
             'all'
         );
@@ -35,11 +37,11 @@ class Cache extends Card
         [$cacheKeyInteractions, $keyTime, $keyRunAt] = $this->remember(
             fn () => $this->aggregateTypes(['cache_hit', 'cache_miss'], 'count')
                 ->map(function ($row) {
-                    return (object) [
-                        'key' => $row->key,
-                        'hits' => $row->cache_hit ?? 0,
-                        'misses' => $row->cache_miss ?? 0,
-                    ];
+                    return new CacheStruct(
+                        key: $row->key,
+                        hits: $row->cache_hit ?? 0,
+                        misses: $row->cache_miss ?? 0,
+                    );
                 }),
             'keys'
         );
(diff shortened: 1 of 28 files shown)


taylorotwell:
I wonder if there is a way to make this work without requiring users to list all of those classes in their configuration. Like should we have `Laravel\Pulse\Pulse::serializableClasses()` which returns an array of these classes so you can do:

```php
[
    ...Pulse::serializableClasses(),
]
```

sgilberg:
@taylorotwell Great idea, I just added the suggested method and tested it in my local app. I also feel like this PR could use some additional tests to make sure that no exceptions are encountered if `cache.serializable_classes` is set in the config, but I'm not sure how best to do that.

I'll also note that after opening this PR, I saw conflicting information on whether or not it is actually safe to unserialize `stdClass` objects. It may be that it is actually safe in the context where we are disallowing unserialization of any potentially nested objects (apart from those explicitly allowed)? I'm happy to be corrected here, if my initial understanding of the threat is incorrect.

taylorotwell:
Getting some test failures. I honestly wonder if we could just cache arrays and cast to object _after_ things are retrieved from the cache. 🤔

sgilberg:
@taylorotwell Thanks, I attempted a fix for those test failures (which I wasn't able to replicate locally, maybe something about my environment). I'll note that the phpstan errors were already in the base branch, and I didn't see any additional ones from this PR when I analyzed locally.

Yeah, we could do arrays and cast to objects. Where that gets mildly annoying is where there are collections of the objects, as we'd need to do I guess a `transform()` on those (and some have nested objects, like the Servers one, with the storage directories, and the Usage one with the resolved user object). If we want to give that a try I could open an alternate PR for that.

valorin:
My understanding is that `stdClass` isn't a deserialisation threat for RCEs because it has no magic methods, and as long as `serializable_classes` allowlist is implemented recursively (which `unserialize()` does by default), then it's safe to include in the list. 

That said, this does have other benefits - stdClass allows an attacker to populate it with absolutely anything they want. While this won't allow recursively loading RCEs, it could still be used maliciously by passing through unexpected values that could manipulate the application in some way. It would require a lot more effort to exploit, but it's still a risk. By giving these objects their own classes, it enforces their specific structure and types - locking down the data and limiting unexpected values.

crynobone (approved):

taylorotwell:
On further research @valorin is correct listing stdClass in safe serializable classes isn't an RCE threat.

PR #510, opened 2026-05-20
Fix Pulse ingest failure when database connection has `use_upsert_alias` enabled

JonPurvis:
## Summary
When a MySQL/MariaDB connection has `use_upsert_alias` enabled ([laravel/framework#42053](https://github.com/laravel/framework/pull/42053)), Laravel generates `INSERT … AS laravel_upsert_alias ON DUPLICATE KEY UPDATE` SQL. Pulse’s `DatabaseStorage` still built aggregate update clauses using `values(column)`, which is incompatible with that syntax and causes ingest to fail with an ambiguous column error.

This PR updates the MySQL/MariaDB upsert expressions in `DatabaseStorage` to use row-alias references when `use_upsert_alias` is enabled, while preserving the existing `values()` behaviour when it is disabled.

## Issue
With `use_upsert_alias => true` on the Pulse storage connection, aggregate upserts generate SQL similar to:
```sql
INSERT INTO `pulse_aggregates` (...)
VALUES (...)
AS laravel_upsert_alias
ON DUPLICATE KEY UPDATE `value` = `value` + values(`value`)
```
MySQL rejects this with:
```text
SQLSTATE[23000]: Column 'value' in field list is ambiguous
```
This is because `Pulse::ingest()` wraps storage in `rescue()`, the exception is captured by default and the dashboard appears empty with no obvious error.

PostgreSQL and SQLite are unaffected; they already use `"excluded"."column"` style references.

## Solution
- Add `usesUpsertAlias()`, `mysqlUpsertTableColumn()`, and `mysqlUpsertIncomingColumn()` helpers on `DatabaseStorage`.
- When `use_upsert_alias` is enabled:
  - Existing row: `` `pulse_aggregates`.`column` ``
  - Incoming row: `` `laravel_upsert_alias`.`column` ``
- When disabled (default / legacy):
  - Existing row: `` `column` ``
  - Incoming row: `` values(`column`) ``
Updated methods: `upsertCount`, `upsertMin`, `upsertMax`, `upsertSum`, `upsertAvg`.
## Test plan
- [x] Added `tests/Feature/Storage/MySqlUpsertAliasTest.php` (requires MySQL/MariaDB; skipped on SQLite)
  - Legacy mode (`use_upsert_alias=false`): SQL contains `values()`, ingest succeeds
  - Alias mode (`use_upsert_alias=true`): SQL contains `laravel_upsert_alias`, ingest succeeds
  - All aggregate types (count, min, max, sum, avg) ingest correctly with alias enabled
  - Documents pre-fix failure when alias + `values()` are mixed

## Backward compatibility
No breaking changes. Applications without `use_upsert_alias` continue to use `values()` as before.

### DIFF
--- a/src/Storage/DatabaseStorage.php
+++ b/src/Storage/DatabaseStorage.php
@@ -188,7 +188,7 @@ class DatabaseStorage implements Storage
             ['bucket', 'period', 'type', 'aggregate', 'key_hash'],
             [
                 'value' => match ($driver = $this->connection()->getDriverName()) {
-                    'mariadb', 'mysql' => new Expression('`value` + values(`value`)'),
+                    'mariadb', 'mysql' => new Expression($this->mysqlUpsertTableColumn('value').' + '.$this->mysqlUpsertIncomingColumn('value')),
                     'pgsql', 'sqlite' => new Expression(<<<SQL
                         {$this->wrap('pulse_aggregates.value')} + "excluded"."value"
                         SQL),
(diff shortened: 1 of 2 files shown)


taylorotwell:

I appreciate you taking the time to submit this; however, it appears this contribution may have been primarily AI-generated without careful human review and consideration.

We've found that AI-generated code often doesn't align well with Laravel's conventions, architectural decisions, and the specific context of what we're trying to accomplish with the framework. Quality contributions require thoughtful human insight into the codebase.

If you're interested in contributing to Laravel, I'd encourage you to familiarize yourself with the existing codebase, engage with the community, and submit PRs that reflect your own understanding and careful consideration of the problem you're solving.

PR #529, opened 2026-07-24
[1.x] Remove userinfo credential from slow outgoing requests

crynobone:
fixes #528


src/Pulse.php
```diff
@@ -551,6 +552,19 @@ public static function defaultVendorCacheKeys(): array
         ];
     }
 
+    /**
+     * Normalize the given URL and mask user & password information.
+     */
+    public static function normalizeUrl(string $url): string
```
timacdonald: It would be good to put this on `Pulse\Recorders\SlowOutgoingRequests`, so that it doesn't get exposed on the facade and because it isn't used anywhere else yet.

src/Pulse.php
```diff
@@ -551,6 +552,19 @@ public static function defaultVendorCacheKeys(): array
...
+    {
+        $uri = Uri::new($url);
+
+        return $uri->withUserInfo(
+            ! is_null($uri->getUsername()) ? Str::mask($uri->getUsername(), '*', 0) : null,
+            ! is_null($uri->getPassword()) ? Str::mask($uri->getPassword(), '*', 0) : null,
```
timacdonald: 🍸 Would it be better to only mask the password?

That might make debugging easier, in case a specific user is creating slow requests to a 3rd party service?

PR #530, opened 2026-07-27
[1.x] Use PHPStan 2

crynobone:


.github/workflows/tests.yml
```diff
@@ -159,7 +159,7 @@ jobs:
           composer update --prefer-dist --no-interaction --no-progress --${{ matrix.stability }}
 
       - name: Execute tests
-        run: vendor/bin/pest -vvv
+        run: vendor/bin/pest -vvv --exclude-group=db-upsert-alias
```
crynobone: Also fix tests CI where MySql 5.7 and MariaDB does not suppoet upsert alias.

PR #539, opened 2026-09-17
[BUG] Prevent newly recorded entries from being ingested before fluent configuration

lcmialichi:
### Description

When the Pulse ingest buffer reaches its limit, the current implementation adds the newly created entry to the buffer and *then* checks for overflow. This causes the newly created entry to be ingested immediately—before the caller has the chance to apply fluent configuration methods like `avg()`, `max()`, or `onlyBuckets()`. 

As a result, the entry is sent to storage missing its aggregation settings, since the fluent methods end up mutating the object in memory *after* it has already been flushed.

### What was changed

1. **Order of Operations:** In `record()`, `set()`, and `lazy()`, I swapped the order so we call `ingestWhenOverBufferSize()` **before** pushing the new item to the arrays. 
2. **Buffer Math:** I updated the capacity checks inside `ingestWhenOverBufferSize()` from `>` to `>=`.
   * *Result:* If the buffer limit is 4, adding the 5th item now successfully flushes the previous 4 items, and then safely places the 5th item into the buffer so it can receive its fluent configurations.

### Testing

* **Buffer Tests:** Updated the `can limit the buffer size of entries` and `resolves lazy entries...` tests to account for the new logic, asserting that the overflowing item correctly remains in the buffer (`wantsIngesting()` stays true until the final flush).
* **Fluent Configuration Test:** Updated the `does not ingest the current entry before fluent configuration is applied` test to be bulletproof. It now mocks the `Ingest` contract and **clones** the entries at the exact moment of ingestion. This proves that the entry actually possesses the fluent configurations *when it is dispatched*, preventing false-positive tests caused by PHP's pass-by-reference object mutation.


taylorotwell:

PR #540, opened 2026-09-18
[1.x] Add SQL Server support

elazaroo:
Pulse's first-party storage currently requires MySQL, MariaDB, PostgreSQL or
SQLite. Applications running on SQL Server have to provision a second database
engine only to store Pulse data. This adds `sqlsrv` to the supported drivers.

SQL Server support was attempted twice before, in #329 and #414. Both stalled
on the same problems, which are addressed here:

**Upserts.** Laravel's SQL Server grammar compiles `upsert()` to a `MERGE`
statement and names the source `laravel_source`, so the aggregate expressions
mirror the PostgreSQL ones with that alias. `least()`/`greatest()` are only
available from SQL Server 2022 onwards, so `iif()` is used instead. The suite
passes with the database at compatibility level 150 (SQL Server 2019).

**Type conversion in the MERGE**, which is what #414 got stuck on. Laravel
binds PHP integers as `PDO::PARAM_INT` and floats as `PDO::PARAM_STR`. SQL
Server resolves the type of each column in a table value constructor by
precedence, and integers outrank strings, so a `value` column mixing both made
the server try to convert `'233.33333333333'` to an integer. Aggregate values
are now normalised before the upsert. The failure was also hard to diagnose:
the aborted transaction made the rollback to the savepoint fail, and that
second exception masked the original one.

**Parameter limit.** SQL Server accepts at most 2,100 parameters per request,
and with the default chunk size of 1,000 records a single ingest of a few
hundred entries exceeded it. As Pulse handles storage exceptions silently, the
data was simply lost. Records are now chunked according to their number of
columns on SQL Server, staying below the limit, and a test storing more records
than fit in a single statement covers it on every driver.

Two smaller changes: the key hash is computed in PHP, reusing the path that
already exists for SQLite, since SQL Server has no MD5 function usable in a
computed column; and `then true` in the type aggregation has been replaced
with `then 1`, which every supported driver accepts.

Tests: a `mssql` job is added to the workflow, running against SQL Server 2022.
The SQL Server connection is configured to preserve numeric types; the tests
that inspect the generated queries also accept `MERGE` statements; and users
created with an explicit primary key toggle `IDENTITY_INSERT`.

The full suite was run against MySQL 5.7, MySQL 8, MariaDB 10, PostgreSQL 14,
SQL Server 2022 and SQLite, with the same results on every driver.


taylorotwell:

elazaroo:
Understood, thanks for taking a look.

Released as a package, as suggested: https://github.com/elazaroo/pulse-sqlsrv

Install with `composer require elazaroo/pulse-sqlsrv.`

It is an add-on rather than a fork: Pulse is installed as usual and the package only swaps the storage implementation when the Pulse connection is `sqlsrv`, overriding the handful of methods that depend on the SQL dialect. Everything else stays on Pulse's own code.

Tested against Laravel 12 and 13, Pulse 1.8, and SQL Server 2019 and 2022.

Leaving it here in case anyone lands on this PR looking for the same thing.

